PR #47, opened 2020-12-05
doc: add asf.yaml .

### LINKED ISSUE #45: add asf.yaml
like https://github.com/apache/apisix/blob/master/.asf.yaml

membphis:
fix #45

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -1,0 +1,41 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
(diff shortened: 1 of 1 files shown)


.asf.yaml
```diff
@@ -0,0 +1,41 @@
...
+      - loadbalancing
+      - apisix
+      - devops
+
+    enabled_merge_buttons:
+      squash:  true
+      merge:   false
+      rebase:  false
+
+    notifications:
```
gxthrj, line 38: `notifications` should be in the first level of yaml

membphis: why? 

`APISIX` AND `APISIX Dashboard`, they are using the same way.

https://github.com/apache/apisix/blob/master/.asf.yaml#L45-L48
https://github.com/apache/apisix-dashboard/blob/master/.asf.yaml#L33-L36

membphis: image

wow @moonming do you know this?

gxthrj (approved):

PR #49, opened 2020-12-05
fix: make labels in asf lowcase

gxthrj:


moonming (approved):

moonming:
There is also a very important point,which is the problem that other ingress open source projects cannot solve.

PR #66, opened 2020-12-08
chore: introduce a CLI interface and a Makefile.

tokers:
The CLI interface is not in use now.


cmd/cmd.go
```diff
@@ -0,0 +1,19 @@
...
+func NewAPISIXIngressControllerCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "apisix-ingress-controller [command]",
+		Long:    "Yet another Ingress controller for Kubernetes using Apache APISIX as the high performance reverse proxy. Please note that all flags in this command line is not in use for now, but will be enabled in the near future.",
+		Version: "", // TODO: fill the version info.
+	}
+
+	cmd.AddCommand(ingress.NewIngressCommand())
+	return cmd
+}
```
gxthrj, line 19: EOL

cmd/ingress/ingress.go
```diff
@@ -0,0 +1,75 @@
...
+
+	// TODO: Uncomment these lines.
+	// cmd.PersistentFlags().StringVar(&configPath, "config-path", "", "file path for the configuration of apisix-ingress-controller")
+	// cmd.PersistentFlags().StringVar(&conf.Kubeconfig, "kubeconfig", "", "Kubernetes configuration file (by default in-cluster configuration will be used)")
+	// cmd.PersistentFlags().StringSliceVar(&conf.Etcd.Endpoints, "etcd-endpoints", nil, "etcd endpoints")
+	// cmd.PersistentFlags().StringVar(&conf.APISIX.BaseURL, "apisix-base-url", "", "the base URL for APISIX instance")
+	// cmd.PersistentFlags().StringVar(&conf.SyslogServer, "syslog-server", "", "syslog server address")
+
+	return cmd
+}
```
gxthrj, line 75: ditto

cmd/cmd.go
tokers: It doesn't show the "No new line at end of file"

build.sh
```diff
@@ -7,5 +7,5 @@ sed -i -e "s%#syslogAddress#%`echo $SYSLOG_HOST`%g" ${pwd}/conf.json
 sed -i -e "s%#apisixBaseUrl#%`echo $APISIX_BASE_URL`%g" ${pwd}/conf.json
 
 cd /root/ingress-controller
-exec ./ingress-controller
+exec ./ingress-controller ingress
```
membphis, line 10: why we need the argument `ingress`?

membphis (changes requested):
missing test cases

build.sh
tokers: We put all core logic about "ingress controller" in the sub command ingress, and we can extend this tool by adding others sub command, for instance, some debugging tools to fetch its internal states, or adding the service apis support by the "gateway" sub command. So using sub commands is more flexible.

tokers:
> missing test cases

It's not easy to add test cases now, since the config is so tricky, then program cannot run unless we modify the configuration file manually. I will add test cases after we refactoring the configuration module.

tokers:
I have created an issue https://github.com/apache/apisix-ingress-controller/issues/70 to track the test cases.

gxthrj (approved):

membphis (commented):
LGTM, add more E2E test in CI

membphis (approved):

PR #77, opened 2020-12-10
chore: config module.

tokers:
Re designed a config module, supporting initializing from command line optoins and files.
Now this module is not in use. Will be used in the near future.

### DIFF
--- a/cmd/ingress/ingress.go
+++ b/cmd/ingress/ingress.go
@@ -111,7 +111,7 @@ func NewIngressCommand() *cobra.Command {
 	cmd.PersistentFlags().StringVar(&cfg.HTTPListen, "http-listen", ":8080", "the HTTP Server listen address")
 	cmd.PersistentFlags().BoolVar(&cfg.EnableProfiling, "enable-profiling", true, "enable profiling via web interface host:port/debug/pprof")
 	cmd.PersistentFlags().StringVar(&cfg.Kubernetes.Kubeconfig, "kubeconfig", "", "Kubernetes configuration file (by default in-cluster configuration will be used)")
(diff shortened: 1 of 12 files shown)


cmd/ingress/ingress.go
```diff
@@ -71,19 +98,22 @@ func NewIngressCommand() *cobra.Command {
 			}()
 
 			router := pkg.Route()
-			err := http.ListenAndServe(":8080", router)
+			err = http.ListenAndServe(":8080", router)
```
gxthrj, line 101: The `:8080` need to use `config.HTTPListen` instead.

pkg/config/config.go
```diff
@@ -0,0 +1,102 @@
...
+	"io/ioutil"
+	"strings"
+	"time"
+
+	"github.com/api7/ingress-controller/pkg/types"
+	"gopkg.in/yaml.v2"
+)
+
+const (
+	_minimalResyncInterval = time.Second
```
gxthrj, line 29: The Frequency  mini interval of resync 1 per second is too high, how about 1min, the same as default interval.

tokers: I think 1min is too large, what about 30s.

cmd/ingress/ingress.go
tokers: Sure, will be changed in next PR :).

This PR will not introduce any codes that use the new config.

gxthrj: OK, understanding.

pkg/config/config.go
gxthrj: ok

tokers: Changed.

membphis (commented):
LGTM

gxthrj (approved):
LGTM

PR #80, opened 2020-12-11
CI: add test ci

gxthrj:


.github/workflows/test-ci.yml
```diff
@@ -0,0 +1,23 @@
...
+    steps:
+    - uses: actions/checkout@v2
+    - name: setup go
+      uses: actions/setup-go@v1
+      with:
+        go-version: '1.13'
+    - name: run test
+      working-directory: ./
+      run: |
+        go test ./...
```
tokers, line 23: Please check the editor, always add the linebreak.

pkg/ingress/apisix/upstream_test.go
```diff
@@ -94,6 +97,7 @@ kind: ApisixUpstream
 apiVersion: apisix.apache.org/v1
 metadata:
   name: httpserver
+  namespace: cloud
```
tokers, line 100: Still use cloud namespace? we should use ingress-apisix.

gxthrj: This is a test-case

.github/workflows/test-ci.yml
gxthrj: Done

pkg/ingress/apisix/upstream_test.go
tokers: OK

.github/workflows/test-ci.yml
```diff
@@ -0,0 +1,23 @@
+name: test-ci
```
tokers, line 1: the name test-ci is too generic, actually it's just unit test, we will have e2e tests in the future, so `unit-test.yaml` is better.

gxthrj: ok, use unit-test-ci.yml

tokers:
LGTM

tokers (approved):

membphis:
nice

PR #81, opened 2020-12-11
fix: skip 3 frames to see the caller of log methods

tokers:


gxthrj (approved):

gxthrj:
It seems that we have a lot of encapsulation levels.

tokers:
> It seems that we have a lot of encapsulation levels.

Just 3 stack frames need to be skipped:

```
zapcore.Write
logger.write
logger.Info
caller
```

PR #82, opened 2020-12-11
feat: install signal handler

tokers:


membphis (changes requested):
test cases ^_^

cmd/ingress/ingress.go
```diff
@@ -39,6 +41,15 @@ func dief(template string, args ...interface{}) {
 	os.Exit(1)
 }
 
+func waitForSignal(stopCh chan struct{}) {
+	sigCh := make(chan os.Signal, 1)
+	signal.Notify(sigCh, syscall.SIGINT, syscall.SIGTERM)
```
gxthrj, line 46: We can add `SIGABRT` and `SIGKILL`, too

tokers: Processes cannot prevent `SIGKILL`.

I think we shouldn't change the behavior of these two.

gxthrj: May be `SIGKILL` only trigged  by people, but `SIGABRT` can be trigged by Kubernetes.

tokers: In which case Kubernetes will send `SIGABRT`? As far as i know, it only sends `SIGTERM` and `SIGKILL`.

gxthrj: It seems that I made a mistake. 
`SIGABRT` could be sent manually by people to view the stack trace.
`SIGKILL ` will be sent by Kubernetes when container is still running after the grace period.

gxthrj: I think we still need to add `SIGABRT` and `SIGKILL `.

tokers: Behavior of SIGKILL cannot be changed, it's kernel's limitation.

I cannot understand why we should capture SIGABRT only because people may send it, if so, wo have to capture all signals since people may send them as well.

membphis (approved):

cmd/ingress/ingress.go
gxthrj: > Behavior of SIGKILL cannot be changed, it's kernel's limitation.

Capture signals for logging, `SIGKILL` would be sent by Kubernetes. why not to capture?

gxthrj: OK，got it

gxthrj (approved):

PR #83, opened 2020-12-11
chore: api server module

tokers:


Firstsawyou:
Need to resolve conflicts.

pkg/api/server.go
```diff
@@ -0,0 +1,64 @@
...
+
+// Run launches the API Server.
+func (srv *Server) Run(stopCh <-chan struct{}) error {
+	go func() {
+		<-stopCh
+		if err := srv.httpListener.Close(); err != nil {
+			log.Errorf("failed to close http listener: %s", err)
+		}
+	}()
+	if err := srv.router.RunListener(srv.httpListener); err != nil && !types.IsUseOfClosedNetConnErr(err) {
```
gxthrj, line 59: should we set some timeout for request?

pkg/api/router/router.go
```diff
@@ -0,0 +1,40 @@
...
+	Status string `json:"status"`
+}
+
+func mountHealthz(r *gin.Engine) {
+	r.GET("/healthz", healthz)
+	r.GET("/apisix/healthz", healthz)
+}
+
+func healthz(c *gin.Context) {
+	c.AbortWithStatusJSON(http.StatusOK, healthzResponse{Status: "ok"})
```
gxthrj, line 34: We had better end with `return`.

pkg/api/server.go
tokers: Currently only a health probe, we can add timeout when we have more APIs.

pkg/api/router/router.go
tokers: Why?

tokers:
> Need to resolve conflicts.

Resolved.

pkg/api/router/router.go
gxthrj: `Abort*`methods does not return directly like its semantics.
If there is logic behind abort, it will continue to execute, so it is recommended to add return and explicitly inform that it will not execute further, although `healthz` is only one line of code :)

tokers: OK, i thought what you said is we should add return statement inside a function without return value.

tokers: Added.

gxthrj:
Need to resolve conflicts

tokers:
> Need to resolve conflicts

Solved.

gxthrj (approved):
LGTM

PR #98, opened 2020-12-15
chore: all modules use new config

tokers:


tokers:
Test case for `kube. BuildRestConfig` will be added after we use terratest.

pkg/ingress/controller/apisix_service.go
```diff
@@ -16,8 +16,8 @@ package controller
 
 import (
 	"fmt"
-	"github.com/api7/ingress-controller/pkg/ingress/apisix"
-	"github.com/golang/glog"
+	"time"
```
gxthrj, line 19: The effect of this `time` was not found in this change.

pkg/ingress/controller/apisix_upstream.go
```diff
@@ -16,9 +16,8 @@ package controller
 
 import (
 	"fmt"
-	"github.com/api7/ingress-controller/pkg/ingress/apisix"
-	"github.com/api7/ingress-controller/pkg/ingress/endpoint"
-	"github.com/golang/glog"
+	"time"
```
gxthrj, line 19: ditto

pkg/kube/init.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+
+	apisixKubeClient, err = clientSet.NewForConfig(restConfig)
+	if err != nil {
+		return err
+	}
+	CoreSharedInformerFactory = informers.NewSharedInformerFactory(kubeClient, 0)
+
+	return nil
+}
```
gxthrj, line 61: It is a big change, we need to test if informers can work as before.

pkg/ingress/controller/apisix_service.go
tokers: image

It's used here.

pkg/ingress/controller/apisix_upstream.go
tokers: It's used here
image

pkg/kube/init.go
tokers: Well, it's not easy to write unit test cases now, i will run them in my local minikube.

gxthrj: ok

gxthrj (approved):
LGTM

tokers:
@gxthrj After this PR is merged, the dev docker image should be updated, also, docs and samples should be changed, since the way we configure ingress apisix is changed.

gxthrj:
> @gxthrj After this PR is merged, the dev docker image should be updated, also, docs and samples should be changed, since the way we configure ingress apisix is changed.

Should we update the docs in this PR?

pkg/kube/init.go
tokers: I have run it in my local minikube, it works well.

tokers:
> > @gxthrj After this PR is merged, the dev docker image should be updated, also, docs and samples should be changed, since the way we configure ingress apisix is changed.
> 
> Should we update the docs in this PR?

That's OK.

tokers:
> > > @gxthrj After this PR is merged, the dev docker image should be updated, also, docs and samples should be changed, since the way we configure ingress apisix is changed.
> > 
> > 
> > Should we update the docs in this PR?
> 
> That's OK.

@gxthrj Added.

PR #102, opened 2020-12-16
fix: set base url to seven module

tokers:


tokers:
@gxthrj After merging, the docker image should be rebuilt 😂

gxthrj:
We can add a test case about `baseUrl`.

tokers:
> We can add a test case about `baseUrl`.

We can only add very shallow test cases like checking the value is really set to conf.BaseUrl.

We cannot test it in other cases like testing APISIXRoute, APISIXUpstream or others.

gxthrj (approved):

PR #106, opened 2020-12-17
fix: tls remove failed

gxthrj:
- [x] Fix: TLS remove failed .
- [x] Trigger retry when the nodes in upstream is empty .

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -43,6 +43,8 @@ Note since APISIX have two configuration files, the first is config.yaml, which 
 
 When `apisix-ingress-controller` creates a route with CRD, it checks the `Endpoint` resources in Kubernetes (matched by namespace_name_port). If the corresponding endpoint information is not found, the route will not be created and wait for the next retry.
 
+Tips: The failure caused by empty upstream nodes is a limitation of Apache APISIX, related [issue](https://github.com/apache/apisix/issues/3072)
+
 6. What is the retry rule of `apisix-ingress-controller`?
 
 If an error occurs during the process of `apisix-ingress-controller` parsing CRD and distributing the configuration to APISIX, a retry will be triggered.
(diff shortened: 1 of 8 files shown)


membphis (commented):
no test case?

pkg/ingress/endpoint/ep.go
```diff
@@ -60,5 +60,12 @@ func BuildEps(ns, name string, port int) []*v1.Node {
 			}
 		}
 	}
+	if len(nodes) < 1 {
```
tokers, line 63: That's not a right behavior, if we really cannot collect the endpoint information, we shouldn't push this upstream to APISIX, instead, logging it and just throw it away.

gxthrj: ok, we do synchronization before sync, if there is no node, we will try again.

gxthrj: done

gxthrj:
> no test case?

Added some test case for group field.

docs/FAQ.md
```diff
@@ -38,3 +38,13 @@ This is because CRDs are generally declared in the file system, and Apply to ent
 So far apisix-ingress-controller doesn't support set admin_key for Apache APISIX, so when you deploy your APISIX cluster, admin_key should be removed from config.
 
 Note since APISIX have two configuration files, the first is config.yaml, which contains the user specified configs, the other is config-default.yaml, which has all default items, config items in these two files will be merged. So admin_key in both files should be removed. You can customize these two configuration files and mount them to APISIX deloyment.
+
+5. Failed to create route with `ApisixRoute`?
+
+When `apisix-ingress-controller` creates a route with CRD, it checks the `Endpoint` resources in Kubernetes (matched by namespace_name_port). If the corresponding endpoint information is not found, the route will not be created and wait for the next retry.
```
tokers, line 44: We should note that this is the limitation of APISIX.

gxthrj: Done

tokers (approved):

PR #127, opened 2020-12-23
chore: prometheus metrics handler

tokers:


gxthrj (approved):

pkg/api/router/router.go
```diff
@@ -34,7 +35,16 @@ func healthz(c *gin.Context) {
 	return
 }
 
+func mountMetrics(r *gin.Engine) {
+	r.GET("/metrics", metrics)
```
membphis, line 39: Will it conflict with the user's existing API?

tokers: It is the API of apisix ingress controller itself, users cannot add apis to it.

membphis: got it

PR #131, opened 2020-12-25
chore: refactor conf/ directory

tokers:
* format `conf.json` is legacy.
* rename `config` to `kubeconfig`, which is more semantic.


moonming (approved):

conf/config-default.json
```diff
@@ -0,0 +1,13 @@
...
+  "http_listen": ":8080",
+  "enable_profiling": true,
+  "kubernetes": {
+    "kubeconfig": "",
+    "resync_interval": "60s"
+  },
+  "apisix": {
+    "base_url": "http://127.0.0.1:9080/apisix/admin"
+  }
+}
```
gxthrj, line 13: There is no place to use this file in the code. 
Is the purpose of this file to give users a configuration example ?

tokers: Two example configuration files, one is in `yaml` and another is in `json`.

PR #143, opened 2020-12-29
chore: add some metrics

tokers:
Abstract a metrics collector `metrics.Collector` and implement it with a Prometheus collector.

The collector is created in `cmd/ingress.Run`, which will be moved to the ingress controller after we have an unified ingress controller object.

Metrics are not used except the `Collector.ResetLeader` (which is always true since we don't use the leader election), others will be used during the refactoring.

### DIFF
--- a/cmd/ingress/ingress.go
+++ b/cmd/ingress/ingress.go
@@ -23,6 +23,7 @@ import (
 	"syscall"
 	"time"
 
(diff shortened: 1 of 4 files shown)


gxthrj:
CI failed.

tokers:
Fixed.

gxthrj (approved):

moonming (approved):

PR #158, opened 2021-01-07
chore: version info in Makefile

tokers:


membphis (approved):

membphis:
we can add a script to check the version when we release new version.

here is an example of APISIX: https://github.com/apache/apisix/blob/master/utils/check-version.sh

tokers:
@membphis OK, will check it out.

Issue #186, opened 2021-01-20
The Ginkgo tool is used in the e2e test script, but the pre-installation check is not performed in the makefile script


liodo198592:
### Issue description
The Ginkgo tool is used in the e2e test script, but the pre-installation check is not performed in the makefile script.

![image](https://user-images.githubusercontent.com/12790469/105144797-bbae7180-5b38-11eb-9557-2fe894db8a29.png)

gxthrj:
Thanks @liodo198592 , There needs to be a check on the ginkgo command.

gxthrj:
Is anyone interested in helping fix this issue?

cao19881125:
let me have a try

PR #190, opened 2021-01-21
chore:add ginkgo check before e2e-test(apache #186)

cao19881125:
Fix: https://github.com/apache/apisix-ingress-controller/issues/186

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,7 @@ VERSION ?= 0.2.0
 RELEASE_SRC = apache-apisix-ingress-controller-${VERSION}-src
 IMAGE_TAG ?= "dev"
 
+GINKGO ?= $(shell which ginkgo)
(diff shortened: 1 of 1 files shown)


gxthrj (approved):
LGTM, Great~

Makefile
```diff
@@ -58,12 +59,18 @@ unit-test:
 	go test -cover -coverprofile=coverage.txt ./...
 
 ### e2e-test:         Run e2e test cases (minikube is required)
-e2e-test: build-image-to-minikube
+e2e-test: build-image-to-minikube ginkgo-check
```
tokers, line 62: We may run `ginkgo-check` before `build-image-to-minikube`, it would be het up if the e2e-test fails after the image was pushed to minikube node but the `ginkgo` tool is not found.

PR #199, opened 2021-01-22
chore: generate id instead let APISIX do it

tokers:

- [ ] Bugfix
- [x] New feature provided
- [ ] Improve performance
- [ ] Backport patches

Previously, we don't generate resource id, instead, we resort to APISIX to do it, which is not so convenient when we try to get resource from APISIX (we'd have to list all resources in APISIX and find the target by fullname).

Now, we generate resource id from fullname.


tokers:
@gxthrj

gxthrj:
UT failed

tokers:
@gxthrj Fixed.

membphis (commented):
LGTM

gxthrj (approved):

Issue #209, opened 2021-01-25
request help: separated base chart or merge base chart


Junnplus:
### Issue description

Hi, I see helm chart is now supported, but I have a confusion:
Why base is separated of controller?

It is possible to deploy multiple for ingress controller, but each controller may have a different serviceaccount/rbac, it is inconvenient to maintain multiple releases for separated base.

reference:
- [kong ingress controller](https://github.com/Kong/charts/tree/main/charts/kong)
- [traefik ingress controller](https://github.com/traefik/traefik-helm-chart/tree/master/traefik)

gxthrj:
> It is possible to deploy multiple for ingress controller, but each controller may have a different serviceaccount/rbac, it is inconvenient to maintain multiple releases for separated base.

This is indeed a more convincing scenario. It is a better choice to merge base into ingress-apisix .

tokers:
Well, now think back it's not a good design to separate them, i suggest to merge them.

gxthrj:
OK, Let's wait to other people's thoughts.

Junnplus:
It seems that not many people use this now. I'd like to refactoring this if nobody minds.

tokers:
@Junnplus Feel free to ask any questions here.

PR #213, opened 2021-01-26
refactoring helm chart

Junnplus:
- Related issues
closes https://github.com/apache/apisix-ingress-controller/issues/209

___

merge base and ingress-apisix helm chart


charts/apisix-ingress-controller/templates/configmap.yaml
```diff
@@ -18,24 +18,24 @@ apiVersion: v1
 data:
   config.yaml: |
     # log options
-    log_level: {{ .Values.ingressController.config.logLevel | quote }}
```
Junnplus, line 21: why remove `ingressController` from yaml?
1. Reduce hierarchy
2. Easy to integrated into [apisix-helm-chart](https://github.com/apache/apisix-helm-chart) as subchart.

charts/apisix-ingress-controller/Chart.yaml
```diff
@@ -15,13 +15,14 @@
...
+name: apisix-ingress-controller
 description: Apache APISIX Ingress Controller for Kubernetes
 icon: https://apache.org/logos/res/apisix/apisix.png
 keywords:
   - ingress
   - apisix
   - nginx
+  - crd
 type: application
 version: 0.1.0
```
tokers, line 27: Since 0.2.0 was released, we can change the `version` to `0.2.0`.

charts/apisix-ingress-controller/README.md
```diff
@@ -0,0 +1,44 @@
...
+# limitations under the License.
+#
+-->
+
+# Apache APISIX Ingress Controller Helm Chart
+
+## Prerequisites
+
+- Kubernetes 1.12+
+- [Apche APISIX](https://github.com/apache/apisix#configure-and-installation)
```
tokers, line 25: Typo: Apache.

charts/apisix-ingress-controller/README.md
```diff
@@ -0,0 +1,44 @@
...
+- [Helm v3.0+](https://helm.sh/docs/intro/quickstart/#install-helm)
+
+## Install
+
+Before installing the chart, you must first install [Apche APISIX](https://github.com/apache/apisix#configure-and-installation).
```
tokers, line 30: Ditto.

tokers: Also, you have indicated the dependencies in `Prerequisities`, this statement is redundant.

charts/apisix-ingress-controller/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+To install the chart with release name `apisix-ingress-controller`:
+
+```bash
+helm install apisix-ingress-controller --namespace apisix .
```
tokers, line 35: Better to add a `cd ./charts/apisix-ingress-controller` if the directory in `helm install` is `.`.

tokers: Also, we recommend to use namespace `ingress-apisix`, and it's better to tell users, the target namespace musy be created in advance.

charts/apisix-ingress-controller/README.md
```diff
@@ -0,0 +1,44 @@
...
+```
+
+## Uninstall
+
+To uninstall/delete the `apisix-ingress-controller` release:
+
+```bash
+helm uninstall apisix-ingress-controller
```
tokers, line 43: Here you should also add a `--namespace` specifier, keep consistent with the `install`.

charts/apisix-ingress-controller/templates/configmap.yaml
tokers: Agree, please revert this change.

charts/apisix-ingress-controller/templates/rbac.yaml
```diff
@@ -14,11 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-apiVersion: v1
-kind: ServiceAccount
-metadata:
-  name: {{ .Release.Name }}-serviceaccount
----
+{{- if .Values.rbac.create }}
```
tokers, line 17: ``{{-`` makes the first line inside the if block joint with the last line, which will cause `apiVersion` commented. You should use `{{` here.

charts/apisix-ingress-controller/templates/rbac.yaml
```diff
@@ -157,5 +153,6 @@ roleRef:
   name: {{ .Release.Name }}-clusterrole
 subjects:
   - kind: ServiceAccount
-    name: {{ .Release.Name }}-serviceaccount
+    name: {{ include "apisix-ingress-controller.serviceAccountName" . }}
```
tokers, line 156: The name of service account should be opaque to the administrator, i don't think it's necessary to customize it, {{ .Release.Name }}-serviceaccount is enough.

Junnplus: I don't think so, service account may not be created, user can use existd name.
see [best practices for RBAC](https://helm.sh/docs/chart_best_practices/rbac/).

charts/apisix-ingress-controller/Chart.yaml
Junnplus: version? or appVersion?

charts/apisix-ingress-controller/README.md
Junnplus: > Also, we recommend to use namespace ingress-apisix, and it's better to tell users, the target namespace musy be created in advance.
agree.

charts/apisix-ingress-controller/templates/rbac.yaml
Junnplus: `{{-` is right, `{{` will keep a blank line.
```
# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
# See the License for the specific language governing permissions and
# limitations under the License.
#

apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
```

charts/apisix-ingress-controller/README.md
Junnplus: > Better to add a cd ./charts/apisix-ingress-controller if the directory in helm install is ..

This readme document provides instructions for this chart, so I think use `.` is ok.

charts/apisix-ingress-controller/Chart.yaml
Junnplus: I think you mean `appVersion`, change the `appVersion` to `0.2.0` is better, also need to be modified `image.tag` in values, but `apisix-ingress-controller` has not yet been release 0.2.0 docker image.

tokers: Yes，the appVersion.

charts/apisix-ingress-controller/templates/rbac.yaml
tokers: It's OK to keep the blank line, but it's a syntax error if the `apiVersion` field is commented.

See https://github.com/helm/helm/issues/9225 for the details.

charts/apisix-ingress-controller/templates/rbac.yaml
tokers: Well, we only need  the permission that we need, and these permissions are defined in the matched rbac.yaml. Why people need to use an existing service account? Why we need to let users to configure other service accounts?
 it's an extra burden for an administrator.

charts/apisix-ingress-controller/templates/rbac.yaml
Junnplus: `{{- if .Values.rbac.create }}` and `{{- if .Values.rbac.create -}}` are different.

`{{- if .Values.rbac.create -}}` will cause `apiVersion` commented

charts/apisix-ingress-controller/templates/rbac.yaml
Junnplus: If we hava multi ingress controller, we can use an existing service account, it depends on the user's choice.

`serviceAccount.create` default is True, so users can not concern, it's not an extra burden.

charts/apisix-ingress-controller/templates/rbac.yaml
tokers: OK, good to know that.

charts/apisix-ingress-controller/templates/rbac.yaml
tokers: If we could, expose as few as possible items. The service account is well designed for the permission, it's not easy to let users mimic one, he/she needs to look the one we provided.

For multiple ingress controllers, i don't think it's useful to share the service account, basically these controllers will spread in different namespaces, while service account should be put in the same namespace with the deployment.

charts/apisix-ingress-controller/Chart.yaml
Junnplus: > Yes，i mean the appVersion.

The above explains why not change the `appVersion` to 0.2.0, Can help publish 0.2.0 docker image?

charts/apisix-ingress-controller/templates/rbac.yaml
tokers: I suggest not to do that, we can consider it in the future if the community really has this demand.￼

Junnplus: helm chart of kong/traefik/nginx ingress controller alway provided this future, I think this is a general future.

Do anyone else have any ideas. cc @gxthrj

charts/apisix-ingress-controller/Chart.yaml
tokers: Well, it needs the apache account.

tokers: Just let it go right now. After the 0.2.0 is released, let's change it.

charts/apisix-ingress-controller/templates/rbac.yaml
tokers: I'd like to know some scenes. I prefer the MVP principle, if we don't know whether it's useful, just don't add it.

Junnplus: Actually, we use a unified administrator service account for nginx ingress controller, so hope can custom service account

tokers:
@Junnplus The installation of CRDs (and the ClusterRole) should be optional, if we have multiple Ingress controller Deployments, only the first installation should apply them, you can refer to https://github.com/Kong/charts/blob/main/charts/kong/values.yaml#L346 as a reference.

Junnplus:
@tokers `installCRDs` of kong is for compatibility helm2. this chart require helm 3.0+.

> With the arrival of Helm 3, we removed the old crd-install hooks for a more simple methodology. There is now a special directory called crds that you can create in your chart to hold your CRDs. These CRDs are not templated, but will be installed by default when running a helm install for the chart. If the CRD already exists, it will be skipped with a warning. If you wish to skip the CRD installation step, you can pass the --skip-crds flag.

tokers:
> @tokers `installCRDs` of kong is for compatibility helm2. this chart require helm 3.0+.
> 
> see more in [chart best practices](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/)
> 
> > With the arrival of Helm 3, we removed the old crd-install hooks for a more simple methodology. There is now a special directory called crds that you can create in your chart to hold your CRDs. These CRDs are not templated, but will be installed by default when running a helm install for the chart. If the CRD already exists, it will be skipped with a warning. If you wish to skip the CRD installation step, you can pass the --skip-crds flag.
> 
> if need to be compatible helm2, I can make additional pr for it.

Sorry, i'm not so familiar with Helm, just want to know how to skip installing CRDs in helm 3.0+.

Junnplus:
@tokers Any other questions?

tokers:
@Junnplus LGTM, will need other reviewers to dome double check. Thank you again!

tokers (approved):

membphis (commented):
LGTM, @starsz do you have time to look at this PR?

gxthrj (commented):
Sorry for the late reply, glad to see your discussion.

charts/apisix-ingress-controller/values.yaml
```diff
@@ -0,0 +1,72 @@
...
+  # The name of the ServiceAccount to use.
+  # If not set and create is true, a name is generated using the fullname template
+  name: ~
+
+replicaCount: 1
+
+image:
+  repository: apache/apisix-ingress-controller
+  pullPolicy: IfNotPresent
+  tag: "0.1.0"
```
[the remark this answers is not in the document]
Junnplus: done.

Junnplus:
@gxthrj Can we merge this PR? cc @tokers

tokers:
@Junnplus Yes it can be, the broken e2e case is not related with the changes in this PR. It should be fixed in another PR.

@gxthrj

tokers:
@Junnplus Please update your branch with the newest master, it fixed the broken e2e case, after that, your PR can be merged :)

Junnplus:
ping @tokers

gxthrj (approved):

PR #218, opened 2021-01-29
chore: change ApisixUpstream, use upstreams instead of ports

tokers:
Also, this PR fixed the bug in endpoints controller, which use the container port as the component in upstream name, now we'll refer to the corresponding service, and find the service port, nevertheless, we didn't handle the endpoints deletion, since in that case, the service was deleted, so we cannot find the reference relationship. But that's OK since even we handle the deletion, what we can do is just reset the nodes array to empty in Upstream (so requests response will be `502 Bad Gateway`), while in our case, requests will be sent to the endpoints, and maybe requests are OK if Pods are alive, or requests will get `504 Gateway Timeout`. Maybe we can use service port name to build the upstream name, if this difference is important.

Anyway, delete Service/Endpoints before the ApisixUpstream is not a good way, and always remove ApisixUpstream before the Service and Endpoints.


tokers:
The change is not what we desired now. Will re-implement.

PR #224, opened 2021-02-02
fix: broken e2e test case

tokers:
The image apache/apisix:dev was updated these two days, the status code when upstream is not available was changed from 502 to 503, so the e2e case should also be adjusted.


tokers:
@gxthrj @membphis Please take a look at this PR.

membphis (approved):

PR #191, opened 2021-01-21
fix: add write permission for leases

liodo198592:
…w_clusterrole.yaml

Please answer these questions before submitting a pull request

- Why submit this pull request?
- [x] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches

- Related issues

___
### Bugfix
- Description

- How to fix?

___
### New feature or improvement
- Describe the details and related test reports.

___
### Backport patches
- Why need to backport?

- Source branch

- Related commits and pull requests

- Target branch

PR #235, opened 2021-02-04
chore: add write permission for leases

tokers:
Co-authored-by 周宇 <zhouyu@jsti.com>.


tokers:
duplicated with #191 .

Issue #201, opened 2021-01-25
Support EndpointSlices


tokers:
# What is EndpointSlices

EndpointSlices provide a simple way to track network endpoints within a Kubernetes cluster. They offer a more scalable and extensible alternative to Endpoints.

# Why we need it

Quoted from https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/.

> The Endpoints API has provided a simple and straightforward way of tracking network endpoints in Kubernetes. Unfortunately as Kubernetes clusters and Services have grown to handle and send more traffic to more backend Pods, limitations of that original API became more visible. Most notably, those included challenges with scaling to larger numbers of network endpoints.
>
> Since all network endpoints for a Service were stored in a single Endpoints resource, those resources could get quite large. That affected the performance of Kubernetes components (notably the master control plane) and resulted in significant amounts of network traffic and processing when Endpoints changed. EndpointSlices help you mitigate those issues as well as provide an extensible platform for additional features such as topological routing.

Also, Istio already supported it. But it's not so emergency, just implement it in the future.

PR #252, opened 2021-02-09
feat: add support for Endpointslices

lianghao208:

- [ ] Bugfix
- [X] New feature provided
- [ ] Improve performance
- [ ] Backport patches

- Related issues

#201


lianghao208:
@tokers @membphis PTAL, thanks!

conf/config-default.yaml
```diff
@@ -30,6 +30,7 @@ log_output: "stderr" # the output file path of error log, default is stderr, whe
 http_listen: ":8080"   # the HTTP Server listen address, default is ":8080"
 enable_profiling: true # enable profileing via web interfaces
                        # host:port/debug/pprof, default is true.
+enable_endpointslice: false # enable endpointSlice support, default is false
```
tokers, line 33: Missing a period `.`.

tokers: What about putting this option to `kubernetes` section?

pkg/ingress/controller/controller.go
```diff
@@ -126,20 +123,28 @@ func NewController(cfg *config.Config) (*Controller, error) {
...
 		apisixUpstreamLister:   sharedInformerFactory.Apisix().V1().ApisixUpstreams().Lister(),
 	}
-	c.translator = kube.NewTranslator(&kube.TranslatorOptions{
-		EndpointsLister:      c.epLister,
-		ServiceLister:        c.svcLister,
-		ApisixUpstreamLister: c.apisixUpstreamLister,
-	})
 
-	c.endpointsController = c.newEndpointsController()
+	if c.cfg.EnableEndpointSlice {
```
tokers, line 132: We should whether the running Kubernetes cluster supports the EndpointsSlice resource, if not, Ingress controller shouldn't launch.

pkg/ingress/controller/endpointslice.go
```diff
@@ -0,0 +1,251 @@
...
+			log.Errorf("port %s in endpointslice %s/%s but not in service", *port.Name, ep.Namespace, ep.Name)
+			continue
+		}
+		nodes, err := c.controller.translator.TranslateUpstreamNodesFromEndpointSlice(ep, svcPort)
+		if err != nil {
+			log.Errorw("failed to translate upstream nodes",
+				zap.Error(err),
+				zap.Any("endpointslice", ep),
+				zap.Int32("port", svcPort),
+			)
```
tokers, line 120: It seems we should return `err` here since error occurred. The same logic in endpoints controller should also be changed.

pkg/kube/translator.go
```diff
@@ -18,6 +18,7 @@ import (
 	"fmt"
 
 	corev1 "k8s.io/api/core/v1"
+	discoveryvibeta1 "k8s.io/api/discovery/v1beta1"
```
tokers, line 21: Typo name, should be `discoveryv1beta1 `.

pkg/kube/translator.go
```diff
@@ -56,7 +60,8 @@ type Translator interface {
 // TranslatorOptions contains options to help Translator
 // work well.
 type TranslatorOptions struct {
-	EndpointsLister      listerscorev1.EndpointsLister
+	// EndpointMode decides what source to use to get endpoint information
```
tokers, line 63: what => which, it's object clauses.

pkg/kube/translator.go
```diff
@@ -120,17 +136,42 @@ func (t *translator) TranslateUpstreamConfig(au *configv1.ApisixUpstreamConfig)
 }
 
 func (t *translator) TranslateUpstream(namespace, name string, port int32) (*apisixv1.Upstream, error) {
-	endpoints, err := t.EndpointsLister.Endpoints(namespace).Get(name)
-	if err != nil {
+	var endpoints *corev1.Endpoints
+	var endpointSlice *discoveryvibeta1.EndpointSlice
+	var nodes []apisixv1.UpstreamNode
+	var err error
```
tokers, line 142: Use `var` block instead of multiple single line `var` statements.
```suggestion
	var (
	    endpoints *corev1.Endpoints
	    endpointSlice *discoveryvibeta1.EndpointSlice
	    nodes []apisixv1.UpstreamNode
	    err error
	)
```

pkg/kube/translator.go
```diff
@@ -120,17 +136,42 @@ func (t *translator) TranslateUpstreamConfig(au *configv1.ApisixUpstreamConfig)
...
+			return nil, &translateError{
+				field:  "endpoints",
+				reason: err.Error(),
+			}
+		}
+		nodes, err = t.TranslateUpstreamNodes(endpoints, port)
+		if err != nil {
+			return nil, err
+		}
+	default:
```
tokers, line 168: Just panic is OK, since we never should have other EndpointMode value.

pkg/ingress/controller/controller.go
```diff
@@ -254,7 +259,7 @@ func (c *Controller) run(ctx context.Context) {
 	}
 
 	c.goAttach(func() {
-		c.epInformer.Run(ctx.Done())
+		c.endpointsController.getInformer().Run(ctx.Done())
```
tokers, line 262: I'm not sure whether I have the correct understanding of this `endpointsController`, but seems it doesn't have big advantages here to abstract it except we can write less `if` and `else` branches.

pkg/ingress/controller/controller.go
lianghao208: refer to: https://kubernetes.io/docs/tasks/administer-cluster/enabling-endpointslices/
To determine whether Kubernetes cluster supports the EndpointsSlice resource, some checks needed to be done.
- kubernetes version
- whether endpointslice feature gate is enabled

tokers:
@lianghao208 PR is good to me, but we need some e2e cases to cover it.

tokers:
@lianghao208 Any updates?

lianghao208:
@tokers 
I haven't figured out a proper way to add e2e test case yet . To enable and disable endpointSlice feature, it needs to run on different versions of minikube, which needs to modify the github workflow, any suggestions?

tokers:
> @tokers
> I haven't figured out a proper way to add e2e test case yet . To enable and disable endpointSlice feature, it needs to run on different versions of minikube, which needs to modify the github workflow, any suggestions?

Just add another workflow with different minikube version, own test cases about endpointSlices will be run under this minikube, while in the normal environment, cases about endpointSlices should be skipped, you can check out the ginkgo help documents about how to skip or focus on specific test cases.

gxthrj:
Need to due with some Conflicting files. @lianghao208

tokers:
Ping @lianghao208

gxthrj:
Close stale PR, feel free to reopen.

Issue #221, opened 2021-02-01
request help:  SSL need to be updated when secret has been modified.


gxthrj:
### Issue description
SSL need to be updated when secret has been modified.

gxthrj:
Close this issue for PR has been merged.

PR #284, opened 2021-03-07
feat: support secret controller 

gxthrj:

- [ ] Bugfix
- [x] New feature provided
- [ ] Improve performance
- [ ] Backport patches

- Related issues
https://github.com/apache/apisix-ingress-controller/issues/221
___


pkg/ingress/controller/secret.go
```diff
@@ -0,0 +1,203 @@
...
+		log.Errorf("found secret object with bad namespace/name: %s, ignore it", err)
+		return
+	}
+	if !c.controller.namespaceWatching(key) {
+		return
+	}
+
+	c.workqueue.AddRateLimited(&types.Event{
+		Type:   types.EventAdd,
+		Object: obj,
```
tokers, line 156: It would be better to pass `key` as the object, because there is a time window between we put the event into queue and we process it, in such period, the object mapped by `key` might be changed again in theory.

gxthrj: I add Key field in `types.Event`, keep  the original meaning of `Object`, because we do need some object information, when data is not available or an error is reported based on the key.

tokers: If data is not available, why we should still process on it? It always means the Kubernetes Control Plane is in trouble, or the networking is partitioned, or the object was deleted, in such a case, just return error and use the retry mechanism is OK.

And, I didn't see some other object fields referred.

gxthrj: Yes, we have verified, if Data not available, will return error.
But we also need to record some message such as resouceVersion, which is in the object.

gxthrj: Have changed obj to key.

tokers (approved):

tokers:
@gxthrj PR's look good to me, but we should add more descriptions about `ApisixTLS` to document, both the typical usage and API references should be recorded, I think we can create an issue to do it.

PR #311, opened 2021-03-23
fix: workqueue need to retry when failed

gxthrj:
Need to retry when failed.

### DIFF
--- a/pkg/ingress/controller/apisix_route.go
+++ b/pkg/ingress/controller/apisix_route.go
@@ -188,15 +188,11 @@ func (c *apisixRouteController) handleSyncErr(obj interface{}, err error) {
 		c.workqueue.Forget(obj)
 		return
 	}
-	if c.workqueue.NumRequeues(obj) < _maxRetries {
-		log.Infow("sync ApisixRoute failed, will retry",
-			zap.Any("object", obj),
-		)
-		c.workqueue.AddRateLimited(obj)
-	} else {
-		c.workqueue.Forget(obj)
-		log.Warnf("drop ApisixRoute %+v out of the queue", obj)
-	}
+	log.Warnw("sync ApisixRoute failed, will retry",
+		zap.Any("object", obj),
+		zap.Error(err),
+	)
+	c.workqueue.AddRateLimited(obj)
 }
 
 func (c *apisixRouteController) onAdd(obj interface{}) {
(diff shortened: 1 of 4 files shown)


tokers:
@gxthrj Please add some explanations to show the purpose you submitted this PR.

gxthrj:
We need to retry when sync CRD  failed.
Use `ItemFastSlowRateLimiter` to keep retrying until it succeeds or a new configuration overwrites the previous outdated configuration.

gxthrj:
@tokers I just found the const `_maxRetries`, It is unreasonable to specify the number of retries because the retries will end in a short time.

gxthrj:
BTW, when trigger retry, we should warn, not `info` level log.
Will modify in this PR, too.

tokers:
> BTW, when trigger retry, we should warn, not `info` level log.
> Will modify in this PR, too.

OK.

tokers:
> We need to retry when sync CRD failed.
> Use `ItemFastSlowRateLimiter` to keep retrying until it succeeds or a new configuration overwrites the previous outdated configuration.

It looks this limiter has an exponential back off? It so, that's would be great.

tokers:
> @tokers I just found the const `_maxRetries`, It is unreasonable to specify the number of retries because the retries will end in a short time.

I referred to the implementation endpoints controller.

pkg/ingress/controller/apisix_route.go
```diff
@@ -188,15 +188,11 @@ func (c *apisixRouteController) handleSyncErr(obj interface{}, err error) {
 		c.workqueue.Forget(obj)
 		return
 	}
-	if c.workqueue.NumRequeues(obj) < _maxRetries {
```
tokers, line 192: See the use pattern in https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/endpoint/endpoints_controller.go#L356.

We just reference it.

gxthrj: I suggest to cancel the maximum number of times limit first, and add it when our admission and crd status are improved. Now we need real-time error reporting to quickly solve the problem.

tokers: I cannot claim a distinct point about this now. Just merge it if it makes some trouble at present.

tokers (approved):

pkg/ingress/controller/apisix_route.go
gxthrj: OK

PR #347, opened 2021-04-09
fix: pass remote_addrs to APISIX

tokers:

- [x] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches

- Related issues

___
### Bugfix
- Description

- How to fix?

___

___
### Backport patches
- Why need to backport?

- Source branch

- Target branch

### DIFF
--- a/pkg/apisix/resource.go
+++ b/pkg/apisix/resource.go
@@ -73,17 +73,18 @@ type item struct {
 }
 
 type routeItem struct {
-	UpstreamId string                 `json:"upstream_id"`
-	ServiceId  string                 `json:"service_id"`
-	Host       string                 `json:"host"`
-	Hosts      []string               `json:"hosts"`
-	URI        string                 `json:"uri"`
-	Vars       [][]v1.StringOrSlice   `json:"vars"`
-	Uris       []string               `json:"uris"`
-	Desc       string                 `json:"desc"`
-	Methods    []string               `json:"methods"`
-	Priority   int                    `json:"priority"`
-	Plugins    map[string]interface{} `json:"plugins"`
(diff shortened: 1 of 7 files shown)


tokers:
@gxthrj Please take a look when you have time.

pkg/apisix/resource.go
```diff
@@ -73,17 +73,18 @@ type item struct {
...
-	Host       string                 `json:"host"`
-	Hosts      []string               `json:"hosts"`
-	URI        string                 `json:"uri"`
-	Vars       [][]v1.StringOrSlice   `json:"vars"`
-	Uris       []string               `json:"uris"`
-	Desc       string                 `json:"desc"`
-	Methods    []string               `json:"methods"`
-	Priority   int                    `json:"priority"`
-	Plugins    map[string]interface{} `json:"plugins"`
+	UpstreamId  string                 `json:"upstream_id"`
```
jbampton, line 76: Generally I like some kind of order in lists, arrays or structures but not sure how this affects the whole system.

So with ordering some people chose an alphabetically ordered list, and others use a custom sort order for maybe the first few items in the list and then alphabetically for the rest of the items in the list.  There are many ways to sort and design data structures. 

Ordering also makes it easier to find existing items in the lists and speeds up adding new items.  
So in this file  there seems to be two `lists` and some other files in this PR have `lists`.
A general code style comment that I always push.

tokers: The order of fields affects the memory alignment but is not so important here, if that's really vital, we can optimize it in the future.

gxthrj (approved):

PR #401, opened 2021-04-22
feat: tcp route translation

tokers:
This Pull Request is based on #400 , so please review and merge it after #400 is merged.

This Pull Request adds the translation for stream_route (`tcp` field from ApisixRoute).

Note the current implementation has a **known** issue:

As APISIX doesn't use `upstream_id` in stream_route, we embed the entire upstream object into stream_route, but here a problem comes in, once the corresponding service's endpoints changed, apisix-ingress-controller cannot fetch the newest change to APISIX as the separate upstream object was not created.

This issue will be fixed once APISIX supports to use `upstream_id` in stream_route.


test/e2e/scaffold/k8s.go
```diff
@@ -213,6 +224,26 @@ func (s *Scaffold) ListApisixRoutes() ([]*v1.Route, error) {
...
+	if err != nil {
+		return nil, err
+	}
+	err = cli.AddCluster(&apisix.ClusterOptions{
+		BaseURL: u.String(),
+	})
+	if err != nil {
+		return nil, err
+	}
+	return cli.Cluster("").StreamRoute().List(context.TODO())
```
gxthrj, line 244: Need to update from `master`, ClusterName need to be specified.

tokers: Got it.

tokers: Fixed.

gxthrj (approved):

PR #411, opened 2021-04-28
feat: add ApisixClusterConfig CRD and global_rule type, client

tokers:
This Pull Request is the first part of support ApisixClusterConfig, in this PR, I added:

* the `ApisixClusterConfig` CRD
* the `ApisixClusterConfig` data structures
* the `GlobalRule` data structure
* the cache, apisix client to operate the `GlobalRule`
* sync `GlobalRule` and `StreamRoute` (it was missed in #401 ) when apisix-ingress-controller is starting


pkg/apisix/cluster.go
```diff
@@ -167,6 +169,16 @@ func (c *cluster) syncCacheOnce() (bool, error) {
 		log.Errorf("failed to list ssl in APISIX: %s", err)
 		return false, err
 	}
+	streamRoutes, err := c.streamRoute.List(context.TODO())
```
gxthrj, line 172: Seems to be a bugfix.

gxthrj (approved):

pkg/apisix/cluster.go
tokers: Yes, it was missing but it's OK since stream route is not in use now.

PR #414, opened 2021-04-29
chore: optimize the apisix cluster processing

tokers:
The processing for the apisix cluster was optimized, so it can be easier to implement the multiple clusters support in the future.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -78,8 +78,18 @@ type KubernetesConfig struct {
 
 // APISIXConfig contains all APISIX related config items.
 type APISIXConfig struct {
-	BaseURL string `json:"base_url" yaml:"base_url"`
+	// DefaultClusterName is the name of default cluster.
+	DefaultClusterName string `json:"default_cluster_name"`
+	// DefaultClusterBaseURL is the base url configuration for the default cluster.
+	DefaultClusterBaseURL string `json:"default_cluster_base_url" yaml:"default_cluster_base_url"`
+	// DefaultClusterAdminKey is the admin key for the default cluster.
 	// TODO: Obsolete the plain way to specify admin_key, which is insecure.
+	DefaultClusterAdminKey string `json:"default_cluster_admin_key" yaml:"default_cluster_admin_key"`
+	// BaseURL is same to DefaultClusterBaseURL.
+	// Deprecated: use DefaultClusterBaseURL instead.
+	BaseURL string `json:"base_url" yaml:"base_url"`
+	// AdminKey is same to DefaultClusterAdminKey.
+	// Deprecated: use DefaultClusterAdminKey instead.
 	AdminKey string `json:"admin_key" yaml:"admin_key"`
 }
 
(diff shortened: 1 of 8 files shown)


pkg/ingress/controller.go
```diff
@@ -318,17 +318,17 @@ func (c *Controller) run(ctx context.Context) {
 	c.metricsCollector.ResetLeader(true)
 
 	err := c.apisix.AddCluster(&apisix.ClusterOptions{
-		Name:     "",
-		AdminKey: c.cfg.APISIX.AdminKey,
-		BaseURL:  c.cfg.APISIX.BaseURL,
+		Name:     c.cfg.APISIX.DefaultClusterName,
+		AdminKey: c.cfg.APISIX.DefaultClusterAdminKey,
+		BaseURL:  c.cfg.APISIX.DefaultClusterBaseURL,
```
tao12345666333, line 323: Is this a breaking change?

pkg/config/config.go
```diff
@@ -78,8 +78,18 @@ type KubernetesConfig struct {
...
+	// DefaultClusterName is the name of default cluster.
+	DefaultClusterName string `json:"default_cluster_name"`
+	// DefaultClusterBaseURL is the base url configuration for the default cluster.
+	DefaultClusterBaseURL string `json:"default_cluster_base_url" yaml:"default_cluster_base_url"`
+	// DefaultClusterAdminKey is the admin key for the default cluster.
 	// TODO: Obsolete the plain way to specify admin_key, which is insecure.
+	DefaultClusterAdminKey string `json:"default_cluster_admin_key" yaml:"default_cluster_admin_key"`
+	// BaseURL is same to DefaultClusterBaseURL.
+	// Deprecated: use DefaultClusterBaseURL instead.
+	BaseURL string `json:"base_url" yaml:"base_url"`
```
tao12345666333, line 90: Should we have a deprecation policy?  e.g. when to actually remove it.

pkg/ingress/controller.go
tokers: Not really, we still reserve `BaseURL` and `AdminKey`.

pkg/config/config.go
tokers: What about:

```
// Deprecated: use DefaultClusterBaseURL instead, BaseURL will be removed once v1.0.0 is released.
```

tao12345666333: SGTM.

tokers: Changed.

tokers: Changed.

gxthrj:
I have two questions.
How to configure multi-clusters ?  
How to match cluster `name` and `admin url` and `key`.

tokers:
> I have two questions.
> How to configure multi-clusters ?
> How to match cluster `name` and `admin url` and `key`?

Not in this PR.

gxthrj:
> > I have two questions.
> > How to configure multi-clusters ?
> > How to match cluster `name` and `admin url` and `key`?
> 
> Not in this PR. See the issue description, multiple-clusters will be supported in the future, this PR is just one of the prerequisites.

It is difficult to set up a multi-cluster according to the current configuration. 

The user modified the configuration according to `defaultxxx` this time, and the next time the multi-cluster is actually configured, it will be modified again. 

It is hoped that this modification can finalize the configuration file structure.

gxthrj:
Got it, with `ApisixClusterConfig` CRD .

gxthrj (approved):

Issue #442, opened 2021-05-12
refactor the annotations process


tokers:
The current process for annotations is pathological and the abstraction is not so good. We need to refactor it so the following goals can be achieved.

* Integrate with new annotations handling will be easily
* Handling these annotations when parsing Ingress resources.

PR #443, opened 2021-05-12
chore: refactor the process of annotations

tokers:

This Pull Request refactored the process about annotations, it abstracts two interfaces:

```go
// Extractor encapsulates some auxiliary methods to extract annotations.
type Extractor interface {
	// GetStringAnnotation returns the string value of the target annotation.
	// When the target annoatation is missing, empty string will be given.
	GetStringAnnotation(string) string
	// GetStringsAnnotation returns a string slice which splits the value of target
	// annotation by the comma symbol. When the target annotation is missing, an empty
	// slice (not a nil slice) will be given.
	GetStringsAnnotation(string) []string
	// GetBoolAnnotation returns a boolean value from the given annotation.
	// When value is "true", true will be given, other values will be treated as
	// false.
	GetBoolAnnotation(string) bool
}

// Handler abstracts the behavior so that the apisix-ingress-controller knows
// how to parse some annotations and convert them to APISIX plugins.
type Handler interface {
	// Handle parses the target annotation and converts it to the type-agnostic structure.
	// The return value might be nil since some features have an explicit switch, users should
	// judge whether Handle is failed by the second error value.
	Handle(Extractor) (interface{}, error)
	// PluginName returns a string which indicates the target plugin name in APISIX.
	PluginName() string
}
```

With the help of these two interfaces, each annotation handler can be implemented independently and hide its details. Also, now annotations process is also integrated into Ingress.

- [ ] Bugfix
- [ ] New feature provided
- [x] Improve performance
- [ ] Backport patches

- Related issues
https://github.com/apache/apisix-ingress-controller/issues/442


tokers:
@tao12345666333 @gxthrj

docs/en/latest/concepts/annotations.md
```diff
@@ -0,0 +1,55 @@
...
+
+This annotation controls which methods are accepted, multiple methods join together with `,`.
+
+Default is `"*"`, which means all HTTP methods are accepted.
+
+Allowlist Source Range
+-----------------------
+
+You can specify the allowed client IP addresses or nets by the annotation `k8s.apisix.apache.org/allowlist-source-range`, multiple IP adddresses or nets join together with `,`,
+for instance, `k8s.apisix.apache.org/allowlist-source-range: 10.0.5.0/16,127.0.0.1,192.168.3.98`. Default value is *empty*, which means the sources are not limited.
```
tao12345666333, line 55:
```suggestion
for instance, `k8s.apisix.apache.org/allowlist-source-range: 10.0.5.0/16,127.0.0.1,192.168.3.98`. Default value is *empty*, which means the source IP is unlimited.
```

docs/en/latest/concepts/annotations.md
```diff
@@ -0,0 +1,55 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+-->
+
+This document describes all supported annotations and their functions. You can add these annotations in the [Ingress](https://kubernetes.io/docs/concepts/services-networking/ingress/) resources so that advanced features in [Apache APISIX](https://apisix.apache.org) can be combined into [Ingress](https://kubernetes.io/docs/concepts/services-networking/ingress) resources.
+
+> Note all keys and values of annotations are strings, so boolean value like `true` and `false` should be represented as `"true"` and `"false"`.
+
+Cors Support
```
gxthrj, line 28: CORS Support

gxthrj, line 28: CORS Support

gxthrj (approved):
LGTM, Except  `CORS` need to be capitalized

docs/en/latest/concepts/annotations.md
tokers: Changed.

docs/en/latest/concepts/annotations.md
tokers: Changed.

tokers: Changed.

tao12345666333 (approved):

Issue #468, opened 2021-05-19
bug: report scheme error when event record


gxthrj:
### Issue description
Report scheme error when event record: no kind is registered for the type v2alpha1.ApisixRoute in scheme "pkg/runtime/scheme.go:100"
![image](https://user-images.githubusercontent.com/4413028/118761807-ab6c9000-b8a7-11eb-9bdf-5f3d47db5550.png)

### Environment

* your apisix-ingress-controller version : master;

LuChenjing:
Met the same issue with above screentshot.
apisix version: apisix-2.1-2.noarch
apisix ingress-controller: 0.6.0


apisix ingress config likes:
```yaml
apiVersion: apisix.apache.org/v2alpha1
kind: ApisixRoute
metadata:
  name: grafana-route
  namespace: onest-observer
spec:
  http:
  - name: grafana
    match:
      paths:
      - "/grafana*"
    backend:
     serviceName: grafana-ha-svc
     servicePort: 3000
```

tao12345666333:
@gxthrj  Please set milestone for this issue

/milestone v1.1

tokers:
> @gxthrj  Please set milestone for this issue
> 
> 
> 
> /milestone v1.1

Done.

gxthrj:
This has been fixed in #469 , let`s close this.

PR #469, opened 2021-05-19
fix: event record scheme error

gxthrj:

- [x] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches

- Related issues

___
### Bugfix
- Description
#468 

___
### Backport patches
- Why need to backport?

- Source branch

- Target branch

### DIFF
--- a/pkg/ingress/controller.go
+++ b/pkg/ingress/controller.go
@@ -25,6 +25,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/client-go/kubernetes/scheme"
 	typedcorev1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	listerscorev1 "k8s.io/client-go/listers/core/v1"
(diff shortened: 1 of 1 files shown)


pkg/ingress/controller.go
```diff
@@ -37,6 +38,7 @@ import (
 	"github.com/apache/apisix-ingress-controller/pkg/apisix"
 	"github.com/apache/apisix-ingress-controller/pkg/config"
 	"github.com/apache/apisix-ingress-controller/pkg/kube"
+	apisixScheme "github.com/apache/apisix-ingress-controller/pkg/kube/apisix/client/clientset/versioned/scheme"
```
tokers, line 41:
```suggestion
	apisixscheme "github.com/apache/apisix-ingress-controller/pkg/kube/apisix/client/clientset/versioned/scheme"
```

Basically, all characters are in the small case as the import package name alias.

pkg/ingress/controller.go
```diff
@@ -166,6 +168,7 @@ func NewController(cfg *config.Config) (*Controller, error) {
 	}
 
 	// recorder
+	utilruntime.Must(apisixScheme.AddToScheme(scheme.Scheme))
```
tokers, line 171: I think to check the return `err` from `apisixscheme.AddToScheme` and return once `err` is not `nil` is better, basically, we don't panic the program if internal errors happen in the same context.

gxthrj: The official example I refer to : https://github.com/kubernetes/sample-controller/blob/master/controller.go#L97

pkg/ingress/controller.go
gxthrj: Done

pkg/ingress/controller.go
tokers: OK, not a big deal.

tokers (approved):

tao12345666333 (approved):

membphis (approved):

PR #507, opened 2021-05-27
feat: subset changes in controllers

tokers:


pkg/types/pod.go
```diff
@@ -54,22 +55,24 @@ func NewPodCache() PodCache {
 }
 
 func (p *podCache) Add(pod *corev1.Pod) error {
-	if pod.Status.Phase != corev1.PodRunning {
-		return ErrPodNotRunning
+	ip := pod.Status.PodIP
```
gxthrj, line 58: Pod assigned ip does not mean that it is directly available. If the pod is not in the ready state, it will cause traffic to fail. The logic of health check is required, or the judgment is combined with the endpoint controller.

tokers: It's not the responsibility of `podCache` to do it, also, it doesn't matter as we only use endpoints subset in the ready part of an Endpoints resource. So unready pod won't be referenced.

gxthrj: Unready pod will be set to nodes in upstream, that is the problem.

gxthrj (approved):

spacewander (approved):

pkg/ingress/pod.go
```diff
@@ -17,6 +17,8 @@ package ingress
 import (
 	"context"
 
+	"github.com/apache/apisix-ingress-controller/pkg/types"
```
starsz, line 20: Ditto.

starsz (approved):

pkg/ingress/pod.go
gxthrj: Ditto what？

tao12345666333 (approved):

pkg/ingress/pod.go
tokers: Fixed.

starsz: Why GitHub only show this comment? 🤔

pkg/ingress/endpoint.go
```diff
@@ -18,6 +18,8 @@ import (
 	"context"
 	"time"
 
+	configv1 "github.com/apache/apisix-ingress-controller/pkg/kube/apisix/apis/config/v1"
```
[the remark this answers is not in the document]
tokers: I have changed it.

PR #527, opened 2021-06-03
fix:  add namespace for the ServiceAccount reference in ClusterRoleBinding

gxthrj:


tao12345666333:
I think it is unnecessary.

tokers:
Title is confusing.

gxthrj:
It will got error if there is no namespace here.

tao12345666333 (approved):

tao12345666333:
Because the kind we used here is `ServiceAccount` which is in the namespace scope, the namespace field must be specified here. If we are using `User` or `Group`, we don't need it.

tokers (approved):

starsz (approved):

nic-chen (approved):

PR #549, opened 2021-06-20
feat: support using apisix client separately with cache bypass

Yiyiyimu:
Signed-off-by: yiyiyimu <wosoyoung@gmail.com>Please answer these questions before submitting a pull request

- [ ] Bugfix
- [x] New feature provided
- [ ] Improve performance
- [ ] Backport patches

- Related issues

___

	- For using apisix go client separately without kubernetes environment, we need to add an option to bypass the cache.
	- Also add `GetById` for resources to directly get them with their real id.
___
### Backport patches
- Why need to backport?

- Source branch

- Target branch

### DIFF
--- a/pkg/apisix/cluster.go
+++ b/pkg/apisix/cluster.go
@@ -128,7 +128,7 @@ func newCluster(o *ClusterOptions) (Cluster, error) {
 	c.globalRules = newGlobalRuleClient(c)
 	c.consumer = newConsumerClient(c)
 
-	c.cache, err = cache.NewMemDBCache()
+	c.cache, err = cache.NewMemDBCache(o.BypassCache)
 	if err != nil {
 		return nil, err
 	}
@@ -282,7 +282,7 @@ func (c *cluster) String() string {
 // HasSynced implements Cluster.HasSynced method.
 func (c *cluster) HasSynced(ctx context.Context) error {
 	// if bypass cache no need to sync
-	if !c.bypassCache {
+	if c.bypassCache {
 		return nil
 	}
 
--- a/pkg/apisix/cluster.go
+++ b/pkg/apisix/cluster.go
@@ -68,10 +68,11 @@ var (
 
 // ClusterOptions contains parameters to customize APISIX client.
 type ClusterOptions struct {
-	Name     string
-	AdminKey string
-	BaseURL  string
-	Timeout  time.Duration
+	Name        string
+	AdminKey    string
+	BaseURL     string
+	BypassCache bool
+	Timeout     time.Duration
 }
 
 type cluster struct {
@@ -84,6 +85,7 @@ type cluster struct {
 	cache        cache.Cache
 	cacheSynced  chan struct{}
 	cacheSyncErr error
+	bypassCache  bool
 	route        Route
 	upstream     Upstream
 	ssl          SSL
@@ -117,6 +119,7 @@ func newCluster(o *ClusterOptions) (Cluster, error) {
 		},
 		cacheState:  _cacheSyncing, // default state
 		cacheSynced: make(chan struct{}),
+		bypassCache: o.BypassCache,
 	}
 	c.route = newRouteClient(c)
 	c.upstream = newUpstreamClient(c)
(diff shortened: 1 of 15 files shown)


pkg/apisix/cluster.go
```diff
@@ -276,6 +281,11 @@ func (c *cluster) String() string {
 
 // HasSynced implements Cluster.HasSynced method.
 func (c *cluster) HasSynced(ctx context.Context) error {
+	// if bypass cache no need to sync
+	if !c.bypassCache {
```
tokers, line 285: It should be `if c.bypassCache`.

pkg/apisix/route.go
```diff
@@ -41,16 +41,24 @@ func newRouteClient(c *cluster) Route {
...
+}
+
 // FIXME, currently if caller pass a non-existent resource, the Get always passes
 // through cache.
-func (r *routeClient) Get(ctx context.Context, name string) (*v1.Route, error) {
+func (r *routeClient) get(ctx context.Context, id string, name string) (*v1.Route, error) {
 	log.Debugw("try to look up route",
 		zap.String("name", name),
 		zap.String("url", r.url),
 		zap.String("cluster", "default"),
```
tokers, line 59: Should also print the route id in logs.

pkg/apisix/route.go
```diff
@@ -41,16 +41,24 @@ func newRouteClient(c *cluster) Route {
...
 	)
-	rid := id.GenID(name)
-	route, err := r.cluster.cache.GetRoute(rid)
+	route, err := r.cluster.cache.GetRoute(id)
```
tokers, line 61: It still queries the cache. Does this PR still work in progress?

pkg/apisix/cluster.go
Yiyiyimu: Thanks! fixed

pkg/apisix/route.go
Yiyiyimu: `GetById` also pass id as `name` in `get`. 
So changed `zap.String("name", name)` to `zap.String("name/id", name)`

pkg/apisix/route.go
Yiyiyimu: Fixed. 
BTW, I noticed only consumers are not imported to cache. Is that on purpose?

tokers: Consumers will be saved into cache.

gxthrj:
What is the usage scenario of this PR? Is there any related issue description? 
I am worried that the feature will cause the cache to be inconsistent with the internal state of APISIX.

tokers (commented):
> What is the usage scenario of this PR? Is there any related issue description?
> I am worried that the feature will cause the cache to be inconsistent with the internal state of APISIX.

APISIX Ingress Controller won't use it, this is for other scenarios.

Yiyiyimu:
> What is the usage scenario of this PR? Is there any related issue description?
> I am worried that the feature will cause the cache to be inconsistent with the internal state of APISIX.

Hi @gxthrj currently I'm using it as apisix go client in https://github.com/api7/kong-to-apisix

pkg/apisix/global_rule.go
```diff
@@ -40,33 +40,46 @@ func newGlobalRuleClient(c *cluster) GlobalRule {
...
-			zap.Error(err),
-		)
-	} else {
-		log.Debugw("failed to find global_rule in cache, will try to lookup from APISIX",
-			zap.String("name", name),
-			zap.Error(err),
-		)
+
+	var globalRule *v1.GlobalRule
+	var err error
```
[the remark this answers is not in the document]
Yiyiyimu: Thx! fixed

pkg/apisix/route.go
Yiyiyimu: Got it. Will fixed it in another PR

pkg/apisix/route.go
tokers: Here you don't change.

Yiyiyimu: 🤣 My bad. Fixed

tokers (approved):

tokers (changes requested):
Won't merge as the E2E test was skipped unexpectedly.

Yiyiyimu:
Close and would create another PR for new implementation

PR #601, opened 2021-07-19
feat: implement schema API

fgksgf:

- [ ] Bugfix
- [x] New feature provided
- [ ] Improve performance
- [ ] Backport patches

- Related issues
___

- Implement schema API for getting plugin's schema from APISIX.

### DIFF
--- a/pkg/apisix/apisix.go
+++ b/pkg/apisix/apisix.go
@@ -56,6 +56,8 @@ type Cluster interface {
 	HealthCheck(context.Context) error
 	// Plugin returns a Plugin interface that can operate Plugin resources.
 	Plugin() Plugin
+	// Schema returns a Schema interface that can fetch schema of APISIX objects.
+	Schema() Schema
 }
 
 // Route is the specific client interface to take over the create, update,
@@ -120,7 +122,12 @@ type Consumer interface {
 
 // Plugin is the specific client interface to fetch APISIX Plugin resource.
 type Plugin interface {
-	Get(context.Context, string) (*v1.Schema, error)
+	List(context.Context) ([]string, error)
+}
+
+// Schema is the specific client interface to fetch the schema of APISIX objects.
+type Schema interface {
+	GetPluginSchema(context.Context, string) (*v1.Schema, error)
 }
 
 type apisix struct {
--- a/pkg/apisix/apisix.go
+++ b/pkg/apisix/apisix.go
@@ -54,6 +54,8 @@ type Cluster interface {
 	Consumer() Consumer
 	// HealthCheck checks apisix cluster health in realtime.
 	HealthCheck(context.Context) error
+	// Plugin returns a Plugin interface that can operate Plugin resources.
+	Plugin() Plugin
 }
 
 // Route is the specific client interface to take over the create, update,
@@ -106,7 +108,7 @@ type GlobalRule interface {
 	Update(context.Context, *v1.GlobalRule) (*v1.GlobalRule, error)
 }
 
-// Consumer it the specific client interface to take over the create, update,
+// Consumer is the specific client interface to take over the create, update,
 // list and delete for APISIX Consumer resource.
 type Consumer interface {
 	Get(context.Context, string) (*v1.Consumer, error)
@@ -114,6 +116,11 @@ type Consumer interface {
 	Create(context.Context, *v1.Consumer) (*v1.Consumer, error)
 	Delete(context.Context, *v1.Consumer) error
 	Update(context.Context, *v1.Consumer) (*v1.Consumer, error)
+}
+
+// Plugin is the specific client interface to fetch APISIX Plugin resource.
+type Plugin interface {
+	Get(context.Context, string) (*v1.Schema, error)
 }
 
 type apisix struct {
--- a/pkg/apisix/apisix.go
+++ b/pkg/apisix/apisix.go
@@ -12,6 +12,7 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
+
 package apisix
 
 import (
@@ -26,9 +27,9 @@ type APISIX interface {
 	// Cluster specifies the target cluster to talk.
 	Cluster(string) Cluster
 	// AddCluster adds a new cluster.
-	AddCluster(*ClusterOptions) error
+	AddCluster(context.Context, *ClusterOptions) error
 	// UpdateCluster updates an existing cluster.
-	UpdateCluster(*ClusterOptions) error
+	UpdateCluster(context.Context, *ClusterOptions) error
 	// ListClusters lists all APISIX clusters.
 	ListClusters() []Cluster
 }
@@ -168,14 +169,14 @@ func (c *apisix) ListClusters() []Cluster {
 }
 
 // AddCluster implements APISIX.AddCluster method.
-func (c *apisix) AddCluster(co *ClusterOptions) error {
+func (c *apisix) AddCluster(ctx context.Context, co *ClusterOptions) error {
 	c.mu.Lock()
 	defer c.mu.Unlock()
 	_, ok := c.clusters[co.Name]
 	if ok {
 		return ErrDuplicatedCluster
 	}
-	cluster, err := newCluster(co)
+	cluster, err := newCluster(ctx, co)
 	if err != nil {
 		return err
 	}
@@ -183,14 +184,14 @@ func (c *apisix) AddCluster(co *ClusterOptions) error {
 	return nil
 }
 
-func (c *apisix) UpdateCluster(co *ClusterOptions) error {
+func (c *apisix) UpdateCluster(ctx context.Context, co *ClusterOptions) error {
 	c.mu.Lock()
 	defer c.mu.Unlock()
 	if _, ok := c.clusters[co.Name]; !ok {
 		return ErrClusterNotExist
 	}
 
-	cluster, err := newCluster(co)
+	cluster, err := newCluster(ctx, co)
 	if err != nil {
 		return err
 	}
--- a/pkg/apisix/cache/cache.go
+++ b/pkg/apisix/cache/cache.go
@@ -65,6 +65,8 @@ type Cache interface {
 	ListGlobalRules() ([]*v1.GlobalRule, error)
 	// ListConsumers lists all consumer objects in cache.
 	ListConsumers() ([]*v1.Consumer, error)
+	// ListSchema lists all schema in cache.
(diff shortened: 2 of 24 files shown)


pkg/types/apisix/v1/types.go
```diff
@@ -463,3 +463,23 @@ func ComposeConsumerName(namespace, name string) string {
 
 	return buf.String()
 }
+
+// Schema represents the schema of APISIX objects.
+type Schema struct {
```
gxthrj, line 468: If you changed the types.go, Please use the `make codegen` to genrate code.
`

pkg/apisix/cluster.go
```diff
@@ -550,3 +557,27 @@ func readBody(r io.ReadCloser, url string) string {
 	}
 	return string(data)
 }
+
+// getSchema returns the schema of APISIX object.
+func (c *cluster) getSchema(ctx context.Context, url string) (string, error) {
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
```
gxthrj, line 563: I am not sure if this url is right. Because I can not see the prefix contains `/schema/plugins/xxx`.
For example, if we want to get the schema of plugin `kafka-logger`. We can use this `http://127.0.0.1:9080/apisix/admin/schema/plugins/kafka-logger`.

fgksgf: I found `/apisix/admin/schema/plugins/<plugin_name>` and `/apisix/admin/plugins/<plugin_name>` return the same result.

pkg/types/apisix/v1/types.go
fgksgf: I did it but nothing updates.

pkg/apisix/cache/cache.go
```diff
@@ -35,6 +35,8 @@ type Cache interface {
 	InsertGlobalRule(*v1.GlobalRule) error
 	// InsertConsumer adds or updates consumer to cache.
 	InsertConsumer(*v1.Consumer) error
+	// InsertSchema adds or updates schema to cache.
+	InsertSchema(*v1.Schema) error
```
tokers, line 39: Better to use `InsertPluginSchema`.

pkg/apisix/apisix.go
```diff
@@ -54,6 +54,8 @@ type Cluster interface {
 	Consumer() Consumer
 	// HealthCheck checks apisix cluster health in realtime.
 	HealthCheck(context.Context) error
+	// Plugin returns a Plugin interface that can operate Plugin resources.
+	Plugin() Plugin
```
tokers, line 58: We may use the unified term, This so-called `Plugin` client actually only returns its schema. We may just use the term `PluginSchema`.

pkg/apisix/cache/cache.go
fgksgf: `v1.Schema` is the schema of APISIX objects, including plugin, route, upstream, etc. So `InsertSchema` can be reused.

pkg/apisix/apisix.go
fgksgf: I will add a `Schema` client for getting schema of APISIX objects (including plugin, route, upstream, etc.) and the `Plugin` will be only used to query names of plugins.

tokers: Then the title is confusing.

fgksgf: updated

pkg/apisix/plugin.go
```diff
@@ -0,0 +1,49 @@
...
+func (p *pluginClient) List(ctx context.Context) ([]string, error) {
+	log.Debugw("try to list plugins' names in APISIX",
+		zap.String("cluster", "default"),
+		zap.String("url", p.url),
+	)
+	pluginList, err := p.cluster.getList(ctx, p.url+"/list")
+	if err != nil {
+		log.Errorf("failed to list plugins' names: %s", err)
+		return nil, err
+	}
```
tokers, line 46: add a debug log to show the plugin list.

pkg/apisix/plugin_test.go
```diff
@@ -0,0 +1,104 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package apisix
+
+import (
+	"context"
+	"encoding/json"
+	"github.com/stretchr/testify/assert"
```
tokers, line 20: Please categorize the imports.

pkg/apisix/schema.go
```diff
@@ -0,0 +1,91 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package apisix
+
+import (
+	"context"
+
+	"github.com/apache/apisix-ingress-controller/pkg/apisix/cache"
```
tokers, line 21: Please categorize the imports, one for 3rd parties, and one for packages in current module.

fgksgf: ok, this should be checked in `make lint`, I will update the config of golangci

pkg/apisix/plugin.go
```diff
@@ -17,7 +17,9 @@ package apisix
 
 import (
 	"context"
+
 	"github.com/apache/apisix-ingress-controller/pkg/log"
```
tokers, line 21: Put local packages at the bottom of the import block.

pkg/apisix/cluster_test.go
```diff
@@ -20,6 +20,7 @@ import (
 	"testing"
 
 	v1 "github.com/apache/apisix-ingress-controller/pkg/types/apisix/v1"
+
```
tokers, line 23: Ditto.

pkg/apisix/cluster.go
```diff
@@ -136,6 +138,9 @@ func newCluster(o *ClusterOptions) (Cluster, error) {
 
 	go c.syncCache()
 
+	ticker := time.NewTicker(o.SyncInterval)
```
tokers: Since this ticker is used by `syncSchema`, just created it inside it.

tokers: Also, ticker should be closed explicitly.

pkg/apisix/plugin.go
gxthrj: @fgksgf  If use goland, you can integrate `goimports-reviser`. It works well for me.

pkg/apisix/cluster.go
fgksgf: here should i put the code that closing the ticker?

fgksgf:
It's weird, all unit tests passed on my PC but failed in CI.

pkg/apisix/cluster.go
```diff
@@ -315,52 +317,74 @@ func (c *cluster) HasSynced(ctx context.Context) error {
 	}
 }
 
-// syncSchema syncs schema from APISIX.
+// syncSchema syncs schema from APISIX regularly according to the interval.
+func (c *cluster) syncSchema(interval time.Duration) {
+	sigCh := make(chan os.Signal, 1)
+	signal.Notify(sigCh, syscall.SIGINT, syscall.SIGTERM)
```
tokers, line 323: May not handle the signal notification here.

tokers: Just do it like `syncCache`.

PS: I know the design of `newCluster` is not so elegant, we may add a context for it, but I'd like to optimize this in another PR in the future.

tokers:
@fgksgf Please check out the CI failure problems, especially for the unit test cases.

pkg/apisix/cluster.go
fgksgf: But it is different from `syncCache`, which is executed only once. However, the schema needs to be synced regularly, so where should I close the ticker?

gxthrj:
> It's weird, all unit tests passed on my PC but failed in CI.

UT passed.

pkg/apisix/cluster.go
gxthrj: cc @tokers
I think syncing the schema is a normal task to prevent the schema from changing after the APISIX upgrade, which requires a regular task to run in the background all the time. So this is not the same as `syncCache`.

Do you have any better way to update the schema, if APISIX DP upgrade?

tokers: cache sync is not an one-kick event, it will be executed once the leader changing.

tokers: > I think syncing the schema is a normal task to prevent the schema from changing after the APISIX upgrade, which requires a regular task to run in the background all the time. So this is not the same as syncCache.

This is by no means related to the signal handling stuff. Of course, we can do it periodically. A Golang program should only have one place handling the signals.

tokers: @fgksgf You may tweak the cluster creation logic, and passing the context to it. Here you can listen to the ctx channel.

gxthrj: All right, so this comment is about signal processing.

gxthrj: Signal processing can be unified.

pkg/apisix/cluster.go
```diff
@@ -301,6 +314,74 @@ func (c *cluster) HasSynced(ctx context.Context) error {
...
+		case <-ctx.Done():
+			break
+		}
+	}
+}
+
+// syncSchemaOnce syncs schema from APISIX once.
+// It firstly deletes all the schema in the cache,
+// then queries and inserts to the cache.
+func (c *cluster) syncSchemaOnce() error {
```
tokers, line 339: context should be propagated here.

fgksgf: then how to use it?

tao12345666333:
This PR need rebase.

pkg/apisix/cluster.go
tokers: `context.TODO()` are used in some places, you can replace them with the passed-in context.

pkg/apisix/cluster.go
```diff
@@ -301,6 +314,74 @@ func (c *cluster) HasSynced(ctx context.Context) error {
...
+	for {
+		if err := c.syncSchemaOnce(ctx); err != nil {
+			log.Warnf("failed to sync schema: %s", err)
+		}
+
+		select {
+		case <-ticker.C:
+			continue
+		case <-ctx.Done():
+			break
```
tokers, line 331: Not `break`, it should be `return`, `break` just let the flow jump out the `select` block but still in the `for` block.

tokers (approved):

gxthrj (approved):

Issue #533, opened 2021-06-07
feat: add support for Knative


fhuzero:
Knative is designed for serverless workloads on top of Kubernetes. [Knative Serving](https://knative.dev/docs/serving/) runs scalable containers based on ingress gateway as its network layer. Currently, a few ingress gateway implementations such as [Istio](https://istio.io), [Kong](https://github.com/kong/kong) and [Gloo](https://github.com/solo-io/gloo) are supported. APISIX ingress controller, however, is not among them.

This issue aims at exploring Knative ingress management mechanism and adding support for Knative. In other words, make APISIX ingress controller an alternative Knative network layer to manage ingress traffic.

Any suggestions or comments are welcome! And thanks @Yiyiyimu and @moonming for being mentors!

tokers:
Interesting, welcome to submit proposals!

MBoui:
any news on the PRs?

fhuzero:
Here is a PR I was working on to add support for Knative #606.

tao12345666333:
Thanks for your contribution, would you like to pick it up? @fhuzero

fhuzero:
@tao12345666333 Thanks for the follow-up! So glad to see that the APISIX ingress controller project has grown a lot since when I created this issue.

My schedule is tight right now, but I'm willing to pick it up from where I left if I have time in the future. By the way, I'm curious about the development machine and environment for the project. Any recommendations for setting up the environment? I used to develop on an AWS EC2 instance but it cost a bit.

tao12345666333:
Typically, we use KIND to start a local cluster when developing locally, and deploy 1 replica of APISIX and etcd in it.  That's enough.

In addition, if you don't want to develop locally, you can also choose a cloud environment like GitPod, which is also very convenient

fhuzero:
Thank you for the info!

PR #606, opened 2021-07-24
feat: add support for Knative

fhuzero:
- [ ] Bugfix
- [x] New feature provided
- [ ] Improve performance
- [ ] Backport patches

- Related issues
#533 

___
This PR provides support for Knative so that APISIX Ingress Controller is an alternative to KIngress implementation. It can replace Istio, Gloo, Kourier, or Kong to serve as the network layer of Knative Serving. APISIX Ingress Controller is now aware of status change of Knative ingress, able to parse it, and can translate it to APISIX custom resource like routes and upstreams.

The current supported API version of Knative ingress is _networking/v1alpha1_.

Some key changes are listed as follows.

- Add a general KnativeIngress interface to encapsulate different versions
- Add a KnativeIngressVersion to support different Knative ingress version
- Add KnativeClient to operate Knative builtin resources
- Add KnativeIngressInformer and KnativeIngressLister
- Add knativeIngressController and its related methods such as event hander
- Add function (TranslateKnativeIngress) to translate KnativeIngress to APISIX routes and upstreams


pkg/ingress/knative_ingress.go
```diff
@@ -0,0 +1,299 @@
...
+func (c *knativeIngressController) sync(ctx context.Context, ev *types.Event) error {
+	ingEv := ev.Object.(kube.KnativeIngressEvent)
+	namespace, name, err := cache.SplitMetaNamespaceKey(ingEv.Key)
+	if err != nil {
+		log.Errorf("found knative ingress resource with invalid meta namespace key %s: %s", ingEv.Key, err)
+		return err
+	}
+
+	var ing kube.KnativeIngress
+	switch ingEv.GroupVersion {
```
tokers, line 91: Version judgement can be made in the `onAdd`, `onUpdate` and `onDelete` callbacks.

pkg/kube/translation/knative_ingress.go
```diff
@@ -0,0 +1,149 @@
...
+		hosts := rule.Hosts
+		if rule.HTTP == nil {
+			continue
+		}
+		for j, httpPath := range rule.HTTP.Paths {
+			var (
+				ups *apisixv1.Upstream
+				err error
+			)
+			knativeBackend := knativeSelectSplit(httpPath.Splits)
```
tokers, line 46: Why just select one? Maybe others should be put to the traffic-split plugin.

pkg/kube/translation/knative_ingress.go
```diff
@@ -0,0 +1,149 @@
...
+			servicePort := knativeBackend.ServicePort
+			serviceName := fmt.Sprintf("%s.%s.%s", knativeBackend.ServiceNamespace, knativeBackend.ServiceName,
```
tokers, line 48: I am confused by the service name. Why here we should concatenate the namespace and name and port, but it was used just as the Kubernetes Service Name. I was wondering whether it really works.

pkg/ingress/knative_ingress.go
fhuzero: I followed version judgement in pkg/ingress/ingress.go [here](https://github.com/apache/apisix-ingress-controller/blob/master/pkg/ingress/ingress.go#L93), which judges ingress version in `sync`, instead of `onAdd`, `onUpdate`, and `onDelete`. Could you please tell me differences between them?

pkg/kube/translation/knative_ingress.go
fhuzero: Thanks for pointing out the problem! `serviceName` should only consists of `knativeBackend.ServiceName`. This was fixed in a later commit but I failed to identify it when I splited the huge PR. I'll fix it soon.

pkg/kube/translation/knative_ingress.go
fhuzero: This is an initial version of Knative support, and `knativeSelectSplit` does not implement traffic split logic of KIngress right now. I plan to add KIngress features like traffic-split, append-headers, etc. in later PR. I can also add traffic-split feature in this PR.

pkg/ingress/knative_ingress.go
tokers: Intercept unexpected resources in an earlier phase is better.

pkg/kube/translation/knative_ingress.go
tokers: But it's so mysterious if we select one of them by the percent, I think just pick the first one is OK.

pkg/ingress/knative_ingress.go
fhuzero: Got it! I'll put version judgement in `onAdd`, `onUpdate`, and `onDelete`.

pkg/kube/translation/knative_ingress.go
fhuzero: OK, I'll just pick the first one.

Yiyiyimu:
Hi @fhuzero I think it would be nice if you could also [include the documentations](https://github.com/apache/apisix-ingress-controller/commit/979531b274751a97a7e39892fe0550dbb256281f) in this PR

gxthrj:
@fhuzero  Some conflicts need to be dealt with.

fhuzero:
> @fhuzero Some conflicts need to be dealt with.

Thanks for your suggestion! The main conflict is package version difference between this PR and knative/pkg used in knative conformance test. I have tried to align the version but failed. Once knative/pkg upgrades their dependency, I'll solve conflicts between this PR and main branch.

Issue #621, opened 2021-08-11
bug: should add tag omitempty to v2alpha1's config item backend and backends?


chzhuo:
### Issue description
Should add tag `omitempty` to v2alpha1's config item `backend` and `backends`
https://github.com/apache/apisix-ingress-controller/blob/master/pkg/kube/apisix/apis/config/v2alpha1/types.go#L95-L99

When using `clientset v2alpha1` to create a `ApisixRoute`, if set `backends` not `nil` and set `backend=nil` you will get error: 
```
"spec.http" must validate one and only one schema (oneOf). Found 2 valid alternatives, spec.http.backend: Invalid value: "null": spec.http.backend in body must be of type object: "null"
```
Because there is a validation in CRD that will checking `backend` and `backends`, so only oneOf `backend` and `backends` should marshal to 

### Environment

apisix-ingress-controller:  
github.com/apache/apisix-ingress-controller v1.0.0

k8s:   
Client Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.2", GitCommit:"092fbfbf53427de67cac1e9fa54aaa09a28371d7", GitTreeState:"clean", BuildDate:"2021-06-16T12:59:11Z", GoVersion:"go1.16.5", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.3", GitCommit:"1e11e4a2108024935ecfcb2912226cedeafd99df", GitTreeState:"clean", BuildDate:"2020-10-14T12:41:49Z", GoVersion:"go1.15.2", Compiler:"gc", Platform:"linux/amd64"}

gxthrj:
@chzhuo Please describe your scenario which does not be define backend(s)

chzhuo:
`backend` is deprecated, so i not set `backend`, i only set `backends`.
I using `apisix-ingress-controller` to create a `ApisixRoute` as below,  but get error `spec.http.backend: Invalid value: "null"`

```go
package main

import (
	"context"

	v2alpha1Config "github.com/apache/apisix-ingress-controller/pkg/kube/apisix/apis/config/v2alpha1"
	v2alpha1Client "github.com/apache/apisix-ingress-controller/pkg/kube/apisix/client/clientset/versioned/typed/config/v2alpha1"
	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
	"k8s.io/apimachinery/pkg/util/intstr"
[...]

PR #622, opened 2021-08-11
bugfix: add tag omitempty for the fields in some data structures

chzhuo:
### Bugfix
- Description
 Using `clientset` to create `ApisixRoute` will get error `spec.http.backend: Invalid value: "null"` when only set `backends`
- Related issues:
   #621
- How to fix?
  adding `omitempty` to `not-required` field


tokers (approved):

tokers:
@chzhuo I rewrote the title to make it more semantic.

tao12345666333 (approved):

lgy1027:
请问下我是安装最新的helm，还是说代码中引入的依赖升到最新版本解决这个问题？

PR #679, opened 2021-09-13
Remove volumeMounts when webhook is disabled

fgksgf:

- [x] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches

- Related issues

___
### Bugfix
- Description
It will occur errors when webhook is disabled but the controller pod tried to mount the volume.

- How to fix?
Add a placeholder in deployment template, decide whether to populate the `volumeMount` or not according to whether the webhook is enabled.

### DIFF
--- a/pkg/apisix/schema.go
+++ b/pkg/apisix/schema.go
@@ -63,7 +63,7 @@ func (sc schemaClient) getSchema(ctx context.Context, name string) (*v1.Schema, 
 		)
 	}
 
-	url := sc.url + "/" + name
+	url := sc.url + name
 	content, err := sc.cluster.getSchema(ctx, url)
 	if err != nil {
 		log.Errorw("failed to get schema from APISIX",
--- a/pkg/apisix/schema.go
+++ b/pkg/apisix/schema.go
@@ -17,6 +17,7 @@ package apisix
 
 import (
 	"context"
+	"path"
 
 	"go.uber.org/zap"
 
@@ -33,7 +34,7 @@ type schemaClient struct {
 
 func newSchemaClient(c *cluster) Schema {
 	return &schemaClient{
-		url:     c.baseURL + "/schema/",
+		url:     path.Join(c.baseURL, "schema"),
 		cluster: c,
 	}
 }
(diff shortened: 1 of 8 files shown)


test/e2e/scaffold/ingress.go
```diff
@@ -276,10 +276,7 @@ spec:
             - --apisix-route-version
             - %s
             - --watch-endpointslices
-          volumeMounts:
-           - name: webhook-certs
-             mountPath: /etc/webhook/certs
-             readOnly: true
+          %s
       volumes:
        - name: webhook-certs
```
gxthrj, line 281: Should this also be set  only when enableWebhook=true ？

pkg/apisix/schema.go
```diff
@@ -63,7 +63,7 @@ func (sc schemaClient) getSchema(ctx context.Context, name string) (*v1.Schema,
 		)
 	}
 
-	url := sc.url + "/" + name
+	url := sc.url + name
```
tokers, line 66: Why change this? Maybe path.Join is more appropriate.

tokers (approved):

pkg/apisix/schema.go
fgksgf: Because I checked logs of E2E test and found there is a duplicated `/`:
image

tokers: OK, use `path.Join` is better.

test/e2e/ingress/webhook.go
```diff
@@ -24,7 +24,7 @@ import (
 	"github.com/apache/apisix-ingress-controller/test/e2e/scaffold"
 )
 
-var _ = ginkgo.Describe("Enable webhooks", func() {
+var _ = ginkgo.FDescribe("Enable webhooks", func() {
```
tokers, line 27: Please don't forget to revert his change before this PR merged.

fgksgf: OK, I'm debugging the webhook test case.

gxthrj (approved):

pkg/apisix/schema.go
fgksgf: It's not suitable to use `path.Join`:
image

test/e2e/scaffold/ingress.go
fgksgf: It doesn't seem to affect

Issue #587, opened 2021-07-13
feat: Need to fully compare CRD and  internal objects in APISIX


gxthrj:
### Issue description
The current Ingress controller is highly available through elections, but when all controller instances are shut down, some events of deleting or modifying resources will be lost, which will cause inconsistencies between the declared configuration and the APISIX state, so we need a full comparison mechanism when apisix ingress starting up.

Discussing in https://lists.apache.org/thread.html/r2c73eca7b61200ffd132d2397d01cf20fa41955c781c34f39d915b83%40%3Cdev.apisix.apache.org%3E

fatalc:
Consider migrating to <https://github.com/kubernetes-sigs/controller-runtime> 😊

tao12345666333:
#680 has been merged. Should we close this one? @gxthrj

gxthrj:
Yes, close it now.

PR #680, opened 2021-09-13
feat: add full compare when ingress startup

gxthrj:

- [ ] Bugfix
- [x] New feature provided
- [ ] Improve performance
- [ ] Backport patches

- Related issues
#587

### DIFF
--- a/pkg/ingress/compare.go
+++ b/pkg/ingress/compare.go
@@ -3,6 +3,7 @@ package ingress
 import (
 	"context"
 	"sync"
+	"time"
 
 	"C"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -10,6 +11,9 @@ import (
 	apisix "github.com/apache/apisix-ingress-controller/pkg/types/apisix/v1"
 )
 
+// CompareResources use to compare the object IDs in resources and APISIX
+// Find out the rest of objects in APISIX
+// AND remove them.
 func (c *Controller) CompareResources() {
 	var (
 		routeMapK8S       = new(sync.Map)
@@ -25,6 +29,20 @@ func (c *Controller) CompareResources() {
 		consumerMapA6    = new(sync.Map)
 	)
 	// todo if watchingNamespace == nil
+	if c.watchingNamespace == nil {
+		opts := v1.ListOptions{}
+		// list all apisixroute resources in all namespaces
+		nsList, err := c.kubeClient.Client.CoreV1().Namespaces().List(context.TODO(), opts)
+		if err != nil {
+			panic(err)
+		} else {
+			wns := make(map[string]struct{}, len(nsList.Items))
+			for _, v := range nsList.Items {
+				wns[v.Name] = struct{}{}
+			}
+			c.watchingNamespace = wns
+		}
+	}
 	for ns, _ := range c.watchingNamespace {
 		// ApisixRoute
 		opts := v1.ListOptions{}
@@ -96,18 +114,18 @@ func (c *Controller) CompareResources() {
 	c.listSSLCache(sslMapA6)
 	c.listConsumerCache(consumerMapA6)
 	// 3.compare
-	routeReult := findRedundant(routeMapK8S, routeMapA6)
-	streamRouteReult := findRedundant(streamRouteMapK8S, streamRouteMapA6)
-	upstreamReult := findRedundant(upstreamMapK8S, upstreamMapA6)
-	sslReult := findRedundant(sslMapK8S, sslMapA6)
-	consuemrReult := findRedundant(consumerMapK8S, consumerMapA6)
+	routeReult := findRedundant(routeMapA6, routeMapK8S)
+	streamRouteReult := findRedundant(streamRouteMapA6, streamRouteMapK8S)
+	upstreamReult := findRedundant(upstreamMapA6, upstreamMapK8S)
+	sslReult := findRedundant(sslMapA6, sslMapK8S)
+	consuemrReult := findRedundant(consumerMapA6, consumerMapK8S)
 	// 4.remove from APISIX
 	c.removeRouteFromA6(routeReult)
 	c.removeStreamRouteFromA6(streamRouteReult)
-	c.removeUpstreamFromA6(upstreamReult)
 	c.removeSSLFromA6(sslReult)
 	c.removeConsumerFromA6(consuemrReult)
-
+	time.Sleep(5 * time.Second)
+	c.removeUpstreamFromA6(upstreamReult)
 }
 
 // findRedundant find redundant item which in src and do not in dest
@@ -127,7 +145,10 @@ func (c *Controller) removeConsumerFromA6(consumers *sync.Map) {
 	consumers.Range(func(k, v interface{}) bool {
 		r := &apisix.Consumer{}
 		r.Username = k.(string)
-		c.apisix.Cluster(c.cfg.APISIX.DefaultClusterName).Consumer().Delete(context.TODO(), r)
+		err := c.apisix.Cluster(c.cfg.APISIX.DefaultClusterName).Consumer().Delete(context.TODO(), r)
+		if err != nil {
+			panic(err)
+		}
 		return true
 	})
 }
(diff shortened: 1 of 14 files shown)


utils/kind-with-registry.sh
```diff
@@ -54,7 +54,7 @@ fi
 echo "Registry Host: ${reg_host}"
 
 # create a cluster with the local registry enabled in containerd
-cat <<EOF | kind create cluster --name "${KIND_CLUSTER_NAME}" --config=-
+cat <<EOF | kind create cluster --image kindest/node:v1.21.2 --name "${KIND_CLUSTER_NAME}" --config=-
```
tokers, line 57: Version should be set as configurable.

pkg/ingress/compare.go
```diff
@@ -0,0 +1,271 @@
...
+		sslMapA6         = new(sync.Map)
+		consumerMapA6    = new(sync.Map)
+	)
+	// todo if watchingNamespace == nil
+	if c.watchingNamespace == nil {
+		opts := v1.ListOptions{}
+		// list all apisixroute resources in all namespaces
+		nsList, err := c.kubeClient.Client.CoreV1().Namespaces().List(context.TODO(), opts)
+		if err != nil {
+			panic(err)
```
tokers, line 51: Panic here is not so friendly. We may just retry the comparison, as it's meaningless for the Ingress controller to skip the comparison and go ahead if the communication with Kubernetes is broken.

pkg/ingress/compare.go
```diff
@@ -0,0 +1,271 @@
...
+		streamRouteMapA6 = new(sync.Map)
+		upstreamMapA6    = new(sync.Map)
+		sslMapA6         = new(sync.Map)
+		consumerMapA6    = new(sync.Map)
+	)
+	// todo if watchingNamespace == nil
+	if c.watchingNamespace == nil {
+		opts := v1.ListOptions{}
+		// list all apisixroute resources in all namespaces
+		nsList, err := c.kubeClient.Client.CoreV1().Namespaces().List(context.TODO(), opts)
```
tokers, line 49: We should not use `context.TODO()`, use the context passed in `c.run`, using `context.TODO()` is not controllable, for instance, you're running ingress controller on local and the process of communication with kubernetes is stuck, you can't quit the program by just click <Ctrl-C>.

pkg/ingress/compare.go
```diff
@@ -0,0 +1,271 @@
...
+			}
+			c.watchingNamespace = wns
+		}
+	}
+	for ns := range c.watchingNamespace {
+		// ApisixRoute
+		opts := v1.ListOptions{}
+		retRoutes, err := c.kubeClient.APISIXClient.ApisixV2beta1().ApisixRoutes(ns).List(context.TODO(), opts)
+		if err != nil {
+			panic(err)
```
tokers, line 65: Ditto. We should not panic the program here.

pkg/ingress/compare.go
```diff
@@ -0,0 +1,271 @@
...
+		} else {
+			for _, r := range retRoutes.Items {
+				tc, err := c.translator.TranslateRouteV2beta1NotStrictly(&r)
+				if err != nil {
+					panic(err)
```
tokers, line 70: Ditto.

pkg/ingress/compare.go
```diff
@@ -0,0 +1,271 @@
...
+
+	"C"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	apisix "github.com/apache/apisix-ingress-controller/pkg/types/apisix/v1"
+)
+
+// CompareResources use to compare the object IDs in resources and APISIX
+// Find out the rest of objects in APISIX
+// AND remove them.
```
tokers, line 30: I'm afraid removing users' data that doesn't create by the APISIX Ingress controller is a **quite dangerous** behavior, people will be confused by this behavior and complain about it.

What we can do is warning users the consequence, we don't have permissions to remove users' data, again, TOO DANGEROUS.

pkg/ingress/compare.go
```diff
@@ -0,0 +1,271 @@
...
+		streamRouteMapK8S = new(sync.Map)
+		upstreamMapK8S    = new(sync.Map)
+		sslMapK8S         = new(sync.Map)
+		consumerMapK8S    = new(sync.Map)
+
+		routeMapA6       = new(sync.Map)
+		streamRouteMapA6 = new(sync.Map)
+		upstreamMapA6    = new(sync.Map)
+		sslMapA6         = new(sync.Map)
+		consumerMapA6    = new(sync.Map)
```
tokers, line 43: I didn't see any scenarios that multiple goroutines use these objects, so why to use `sync.Map` here, it's overengineering.

pkg/ingress/compare.go
```diff
@@ -0,0 +1,271 @@
...
+					panic(err)
+				} else {
+					consumerMapK8S.Store(consumer.Username, consumer.Username)
+				}
+			}
+		}
+	}
+
+	// 2.get all cache routes
+	c.listRouteCache(routeMapA6)
```
tokers, line 125: Errors that occurred during the communication with Kubernetes will cause the program to be panicked but were ignored  during the communication  with Apache APISIX, it's wrong.

pkg/ingress/compare.go
```diff
@@ -0,0 +1,271 @@
...
+	streamRouteReult := findRedundant(streamRouteMapA6, streamRouteMapK8S)
+	upstreamReult := findRedundant(upstreamMapA6, upstreamMapK8S)
+	sslReult := findRedundant(sslMapA6, sslMapK8S)
+	consuemrReult := findRedundant(consumerMapA6, consumerMapK8S)
+	// 4.remove from APISIX
+	c.removeRouteFromA6(routeReult)
+	c.removeStreamRouteFromA6(streamRouteReult)
+	c.removeSSLFromA6(sslReult)
+	c.removeConsumerFromA6(consuemrReult)
+	time.Sleep(5 * time.Second)
```
tokers, line 141: Don't reply on a FIXED WINDOW. Use poll mechanism.

utils/kind-with-registry.sh
gxthrj: I added it in my env for installation. This is not the feature of this PR. 
Will add another PR for this.

pkg/ingress/compare.go
gxthrj: As we discuss in the mailing list, Have changed to logging the differents stauts.

pkg/ingress/compare.go
gxthrj: the variable `*A6` has changed to `map[string]string`, but keep `*K8S` using `sync.Map`.

pkg/ingress/compare.go
gxthrj: That is because the cache has been verified before. cc https://github.com/apache/apisix-ingress-controller/blob/master/pkg/ingress/controller.go#L389

tokers:
@gxthrj There are a couple of lint errors, please fix them.

pkg/ingress/compare.go
gxthrj: I think if there is an error when list resources,  it should be panic here.

pkg/ingress/compare.go
gxthrj: Done

pkg/ingress/compare.go
```diff
@@ -0,0 +1,222 @@
...
+import (
+	"context"
+	"sync"
+
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/apache/apisix-ingress-controller/pkg/log"
+)
+
+// CompareResources use to compare the object IDs in resources and APISIX
```
tao12345666333, line 26:
```suggestion
// CompareResources used to compare the object IDs in resources and APISIX
```

pkg/ingress/compare.go
tao12345666333: I also think the panic here is not friendly. WDYT @lingsamuel

gxthrj: I think it needs the same treatment as [here](https://github.com/apache/apisix-ingress-controller/blob/master/cmd/ingress/ingress.go#L123-L126) because it only runs when starting the Ingress controller.

When `controller.NewController(cfg)` failed, it will `os.Exit(1)`, which has the same effect as `panic` .

gxthrj: Or I can return the error which will be captured by the upper layer, and finally call the [dief](https://github.com/apache/apisix-ingress-controller/blob/master/cmd/ingress/ingress.go#L34) method to exit the program?

tao12345666333: > Or I can return the error which will be captured by the upper layer, and finally call the dief method to exit the program?

I prefer this one.

lingsamuel (approved):

tao12345666333 (approved):

PR #723, opened 2021-10-27
Bugfix: add tag omitempty for the fields in this data structure ApisixRouteHTTPMatchExpr

chzhuo:

- [x] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches

- Related issues

___
### Bugfix
- Description
In the ApisixRoute CRD: 
image
So the field `value` and `set` can't appear at the same time
But the golang data structure will be marshal to `value: null` or `set: null`, this will cause CRD checking error

- How to fix?
 Adding tag `omitempty` to this field


tao12345666333 (approved):

tao12345666333:
Thanks!

Issue #690, opened 2021-09-24
bug: Ingress Status Load Balancer IPs Not Being Updated


myoss:
### Environment

* apisix-ingress-controller version: 1.2.0
* your Kubernetes cluster version: ACK 1.20.4-aliyun.1;
* if you run apisix-ingress-controller in Bare-metal environment, also show your OS version (`uname -a`): Linux demo 4.19.91-23.al7.x86_64 #1 SMP Tue Mar 23 18:02:34 CST 2021 x86_64 x86_64 x86_64 GNU/Linux

### Minimal test code / Steps to reproduce the issue

1. ACK 1.20.4-aliyun.1
2. create Ingress
```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: ingress-demo
  namespace: default
spec:
  ingressClassName: apisix
  rules:
    - host: demo.example.com
      http:
        paths:
          - backend:
              service:
                name: svc-demo
                port:
                  number: 80
            path: /
            pathType: Prefix

```
3.

### What's the actual result? (including assertion message & call stack if applicable)
![Snipaste_2021-09-24_18-41-04](https://user-images.githubusercontent.com/6986459/134661969-ec598d35-5359-4126-ad6b-1065212818cd.jpg)

### What's the expected result?
![Snipaste_2021-09-24_18-40-02](https://user-images.githubusercontent.com/6986459/134662520-3089a9c7-0559-4152-b380-a95b7fbd30fe.jpg)

gxthrj:
I think it should be configured in ACK, Perhaps use SLB with it.

myoss:
use nginx-ingress is updated, but apisix-ingress not updated.

tokers:
Thanks for the report, it's an undeveloped feature of ingress, I think we can support it in the next few release cycles.

See:

* https://github.com/projectcontour/contour/blob/main/design/ingress-status-loadbalancer.md
* https://kubernetes.io/docs/reference/kubernetes-api/service-resources/ingress-v1/#IngressStatus

tokers:
We can make some discussions in the mailing list.

myoss:
In a planned way?

tao12345666333:
#740 has been merged. I will close this one. Feel free to reopen it.

@Bonitas you can try to build apisix-ingress-controller using latest code, or wait for v1.4 release.

PR #740, opened 2021-11-08
feat: Fix lb ingress status

tao12345666333:

- [x] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches

- Related issues
#690 
___
### Bugfix
- Description

- How to fix?

___

___
### Backport patches
- Why need to backport?

- Source branch

- Target branch


pkg/ingress/ingress.go
```diff
@@ -184,14 +185,64 @@ func (c *ingressController) sync(ctx context.Context, ev *types.Event) error {
...
 	}
 	log.Warnw("sync ingress failed, will retry",
 		zap.Any("object", obj),
 		zap.Error(err),
 	)
+
+	if errLocal == nil {
+		switch ing.GroupVersion() {
+		case kube.IngressV1:
+			c.controller.recordStatus(ing.V1(), _resourceSynced, err, metav1.ConditionTrue, ing.V1().GetGeneration())
```
gxthrj, line 235:
```suggestion
			c.controller.recordStatus(ing.V1(), _resourceSyncAborted, err, metav1.ConditionTrue, ing.V1().GetGeneration())
```

tokers: This is the case that the `errLocal` is `nil`.

pkg/ingress/status.go
```diff
@@ -163,8 +173,142 @@ func (c *Controller) recordStatus(at interface{}, reason string, err error, stat
...
+	ingressStatusAddress := c.cfg.IngressStatusAddress
+	addrs := []string{}
+
+	// if ingressStatusAddress is specified, it will be used first
+	if len(ingressStatusAddress) > 0 {
+		addrs = append(addrs, ingressStatusAddress...)
+		return addrs, nil
+	}
+
+	namespace, name, err := cache.SplitMetaNamespaceKey(ingressPublishService)
```
tokers, line 249: The separation  can be handled in advance.

pkg/ingress/status.go
```diff
@@ -163,8 +173,142 @@ func (c *Controller) recordStatus(at interface{}, reason string, err error, stat
...
+// ingressPublishAddresses get addressed used to expose Ingress
+func (c *Controller) ingressPublishAddresses() ([]string, error) {
+	ingressPublishService := c.cfg.IngressPublishService
+	ingressStatusAddress := c.cfg.IngressStatusAddress
+	addrs := []string{}
+
+	// if ingressStatusAddress is specified, it will be used first
+	if len(ingressStatusAddress) > 0 {
+		addrs = append(addrs, ingressStatusAddress...)
+		return addrs, nil
```
tokers, line 246: As per the description of this option, the static addresses should only be used if the LB is not ready.

pkg/ingress/status.go
```diff
@@ -163,8 +173,142 @@ func (c *Controller) recordStatus(at interface{}, reason string, err error, stat
...
+	}
+
+}
+
+// ingressLBStatusIPs organizes the available addresses
+func (c *Controller) ingressLBStatusIPs() ([]apiv1.LoadBalancerIngress, error) {
+	lbips := []apiv1.LoadBalancerIngress{}
+	var ips []string
+
+	for {
```
tokers, line 289: Use https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait

pkg/ingress/status.go
tao12345666333: I think that directly specifying the IP will be more direct and will have a higher priority. I will modify the description in the configuration.

pkg/ingress/status.go
tao12345666333: I didn't understand what you mean.

tao12345666333: This specified service is only needed here, I don’t think it needs to be checked or processed in advance.

tokers: Since the namespace is passed from users, we can split it and check the validity when the program is starting, so we can avoid such an error check here and we can abort the program in time (if the format is incorrect).

tao12345666333: But users may not configure Ingress resources.

If the user configures this option, but he only uses custom resources, do we also need to report an error? And if there is any error, it will not affect the actual use of the user.

tao12345666333: Because it is not a parameter required by the real runtime, I don’t want to refuse to provide services to users in advance.

pkg/ingress/status.go
tao12345666333: The implementation of the wait package also uses the for statement, we can optimize it later.

tao12345666333:
I will fix CI ASAP

gxthrj (approved):

tokers (approved):

Issue #741, opened 2021-11-08
bug: The newly created namespace will not be watched


chzhuo:
### Environment

* controller: master 65f7c88193eb6e83b2bb4ca87a981321a99503e5

### Minimal test code / Steps to reproduce the issue
1. start controller
2. create a new namespace
3. namespace will not store to `watchingNamespace`

### The reason of the bug
![image](https://user-images.githubusercontent.com/10919124/140680435-710bf50e-b7af-4027-a871-e44fac82a12b.png)
The method `IsSubsetOf` return `false` when `watchingLabels` is empty

![image](https://user-images.githubusercontent.com/10919124/140680620-65c6386e-23d6-4045-9801-0dfd67edb17e.png)

I think should return `true` when the `s` is empty

tao12345666333:
ping @gxthrj  PTAL

gxthrj:
The reason for the issue is the way to determine if `sync.Map` is empty has wrong. It can not be nil anyway.

gxthrj:
> The method IsSubsetOf return false when watchingLabels is empty
I think should return true when the s is empty

Yes, you are right. Thanks for the report.

tao12345666333:
#742 has been merged, I will close this one. Feel free to reopen it if you have any questions.

PR #742, opened 2021-11-08
fix: ingress do not watching any namespace when namespaceSelector is empty

gxthrj:
…empty

- [x] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches

- Related issues
#741 
___

### DIFF
--- a/pkg/ingress/compare.go
+++ b/pkg/ingress/compare.go
@@ -27,6 +27,8 @@ import (
 // CompareResources used to compare the object IDs in resources and APISIX
 // Find out the rest of objects in APISIX
 // AND warn them in log.
+// This func is NOT concurrency safe.
+// cc https://github.com/apache/apisix-ingress-controller/pull/742#discussion_r757197791
 func (c *Controller) CompareResources(ctx context.Context) error {
 	var (
 		wg                sync.WaitGroup
--- a/pkg/ingress/compare.go
+++ b/pkg/ingress/compare.go
@@ -20,6 +20,7 @@ import (
 
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+	"github.com/apache/apisix-ingress-controller/pkg/api/validation"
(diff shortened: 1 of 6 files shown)


chzhuo:
In the method `initWatchingNamespaceByLabels`, The `watchingNamespace` will be filled .
image

gxthrj:
> In the method `initWatchingNamespaceByLabels`, The `watchingNamespace` will be filled . image

Yes， I just get your point, the newly created namespace will fail to watch.

lingsamuel (approved):

pkg/ingress/compare.go
```diff
@@ -42,7 +43,7 @@ func (c *Controller) CompareResources(ctx context.Context) error {
 		consumerMapA6    = make(map[string]string)
 	)
 	// watchingNamespace == nil means to monitor all namespaces
-	if c.watchingNamespace == nil {
+	if !validation.HasValueInSyncMap(c.watchingNamespace) {
```
tokers, line 46: This check is not safe. The result is only correct at a very specific point-in-time, which means the map might contains some elements after you check it. Please assess the impacts. If necessary, use a mutex.

tao12345666333 (approved):

tao12345666333:
ping @gxthrj  any update?

pkg/ingress/compare.go
gxthrj: I think it is not necessary, because the `sync.Map` in `HasValueInSyncMap ` is concurrency safe, which use `atomic.Value` and `Mutex` inside.

tao12345666333: @tokers I hope we can reach an agreement in this regard.

tokers: OK, I check out the code base I found there is only place use the `CompareResources` and at that time the controllers are not launched.

So let's just add some comments on this method so that once this method is called in the runtime, writers won't make a mistake.

tokers (approved):

Issue #759, opened 2021-11-19
bug: If resource synchronization retry occurs, other events of the same resource will be blocked


gxthrj:
### Issue description
If resource synchronization retry occurs, other events of the same resource will be blocked.

### Environment

* your apisix-ingress-controller version (output of `apisix-ingress-controller version --long`); master

### Minimal test code / Steps to reproduce the issue

1. Define an ApisixRoute resource, including backends that does not exist.
2. The ingress controller will report an error and keep retrying.
3. After waiting for a period of time, create another new ApisixRoute resource and find that the synchronization time is slower.
4. At this time, if you create multiple new ApisixRoute resources, the synchronization time will become slower and slower.

### What's the expected result?
The retry of a resource will not affect the synchronization of other resources of the same type.

tokers:
The maximum retry times should be limited just like what the kube-controller-manager does. We may set the amount to a number like `16`.

tao12345666333:
#760 has been merged. I think we can close this issue.

PR #760, opened 2021-11-19
fix: If resource synchronization retry occurs, other events of the same resource will be blocked.

gxthrj:

- [x] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches

- Related issues
#759 
___
### Bugfix

- How to fix?

This bug is due to a `workqueue` shared under the same resource, and a `ratelimit` mechanism is added to this `workqueue`, but we only need to add the `ratelimit` when retrying fails, and when normal resource changes, we should immediately add the `workqueue` to be processed .

The new k8s event does not need to use the rate limit.

### DIFF
--- a/pkg/ingress/apisix_cluster_config.go
+++ b/pkg/ingress/apisix_cluster_config.go
@@ -207,7 +207,7 @@ func (c *apisixClusterConfigController) onAdd(obj interface{}) {
 		zap.Any("object", obj),
 	)
 
-	c.workqueue.AddRateLimited(&types.Event{
+	c.workqueue.Add(&types.Event{
 		Type:   types.EventAdd,
 		Object: key,
 	})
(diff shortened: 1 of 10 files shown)


tokers:
@gxthrj Could you give some details about the fixup?

tao12345666333 (approved):

gxthrj:
> @gxthrj Could you give some details about the fixup?


This bug is due to a `workqueue` shared under the same resource, and a `ratelimit` mechanism is added to this `workqueue`, but we only need to add the `ratelimit` when retrying fails, and when normal resource changes, we should immediately add the `workqueue` to be processed .

The new k8s event does not need to use the rate limit.

tokers (approved):

crazyMonkey1995:
This fix cannot solve the problem in some scenarios, for example:
Suppose there are two instances of apisix-ingress-controller. When the leader goes down for some reason and another instance becomes the leader, the new leader will block the resource event due to "client-side throttling" during the list resource phase. Because apisix-ingress-controller has not completed the list stage at this time, the control loop will definitely be blocked
image

tao12345666333:
@crazyMonkey1995  Thanks for your report.  We can use #822 for tracking this one.

Issue #769, opened 2021-11-24
bug: Unable to reconnect to apisix, when all ep are deleted under svc of apisix


han6565:
### Issue description
我在k8s上部署了一个apisix，一个apisix-controller，两台配置好后controller健康检查无问题，
当apisix pod被kill后就算pod重新拉起，controller仍然无法重新连接到apisix
但是我后续手动admin访问pod相同地址没有问题
如果有两个或多个apisix pod时只要存在一个活的pod就不会有问题，但是当短时间内没有pod存活就会无法再连接上
是我配置问题，还是本来没有pod能访问时就应该再也无法连接呢

在controller启动同步后，删除service再恢复可以百分之百复现

```
021-11-23T20:51:39+08:00	warn	apisix/cluster.go:452	failed to check health for cluster default: dial tcp 172.24.150.14:9180: connect: connection refused, will retry
2021-11-23T20:51:39+08:00	warn	ingress/controller.go:660	failed to check health for default cluster: timed out waiting for the condition, give up leader
2021-11-23T20:51:39+08:00	info	ingress/endpoint.go:83	endpoints controller exited
2021-11-23T20:51:39+08:00	info	ingress/apisix_tls.go:71	ApisixTls controller exited
2021-11-23T20:51:39+08:00	error	ingress/ingress.go:63	cache sync failed
2021-11-23T20:51:39+08:00	info	ingress/ingress.go:64	ingress controller exited
2021-11-23T20:51:39+08:00	info	ingress/apisix_upstream.go:71	ApisixUpstream controller exited
2021-11-23T20:51:39+08:00	info	ingress/secret.go:76	secret controller exited
2021-11-23T20:51:39+08:00	info	ingress/service.go:61	svc controller exited
2021-11-23T20:51:39+08:00	info	ingress/namespace.go:82	namespace controller exited
2021-11-23T20:51:39+08:00	info	ingress/pod.go:56	pod controller exited
2021-11-23T20:51:39+08:00	info	ingress/apisix_route.go:71	ApisixRoute controller exited
2021-11-23T20:51:39+08:00	info	ingress/apisix_consumer.go:67	ApisixConsumer controller exited
2021-11-23T20:51:39+08:00	info	ingress/apisix_cluster_config.go:69	ApisixClusterConfig controller exited
[...]

PR #770, opened 2021-11-25
feat: implement apisix healthz check

mayocream:
Please answer these questions before submitting a pull request

- Why submit this pull request?
- [ ] Bugfix
- [x] New feature provided
- [ ] Improve performance
- [ ] Backport patches

- Related issues
Fix #567

___
### New feature or improvement
- Implement apisix healthz check, sharing health check state with controller, add unit tests.

Comment if something is wrong, then I'll change it.

PR #774, opened 2021-11-26
fix: delete the cluster object when give up the leadership

tokers:

- [x] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches

- Related issues
#769 
___

In the previous implementation, the cluster object won't be removed if
the controller gives up the leadership, which causes some error state is
cached even if the next leader takes office, this hurts the running.

In this pull request, the cluster is deleted once the leadership is
given up, and the retry-logic will be aborted (for the cluster added in
last term) if the context is canceld.

Signed-off-by: Chao Zhang <tokers@apache.org>


tao12345666333:
CI failed.

tokers:
> CI failed.

The go version on my local is `1.17` and the build directive is different.

tokers:
> > CI failed.
> 
> The go version on my local is `1.17` and the build directive is different.

I use the go/1.16.10 to re-generate some files.

tools.go
```diff
@@ -1,3 +1,4 @@
+//go:build tools
```
tao12345666333, line 1: This is new in 1.17, I want to delete it before we fully switch to v1.17

gxthrj:
better to add e2e for this feature.

tokers:
> better to add e2e for this feature.

First of all, this is not a feature, indeed it's a bugfix, also, it's tough to add cases for it.

tools.go
tokers: Removed.

tao12345666333: You haven't pushed the latest code yet?

tao12345666333:
Maybe we can introduce chaos test later.

If you want to add tests here,
We can first set the number of replicas of apisix to 0 to ensure that all ep are cleaned up.

We may also need to rely on https://github.com/apache/apisix-ingress-controller/pull/770 to confirm that the current cluster is not available

gxthrj:
> > better to add e2e for this feature.
> 
> First of all, this is not a feature, indeed it's a bugfix, also, it's tough to add cases for it.

I think the bugfix also need e2e test , You can restart the APISIX pod during the test, and then check the behavior of the Ingress controller.

tools.go
tao12345666333: ping @tokers

tao12345666333: @tokers  any update?

tokers: I have updated it but I don't have time to validate the e2e case until this weekend.

test/e2e/chaos/chaos.go
```diff
@@ -0,0 +1,71 @@
+package chaos
```
tao12345666333, line 1: @nic-6443 please add license header

tools.go
tao12345666333, line 1:
```suggestion
```
This line should be removed.  @tokers @nic-6443

tao12345666333: Since this is a bug, I hope to merge it into v1.4 as soon as possible, and I will continue to complete this PR. cc @tokers @nic-6443 Thanks for your contribution.

tao12345666333 (approved):
LGTM

gxthrj (approved):

PR #780, opened 2021-12-01
fix: test case param error

lxm:
Signed-off-by: lxm <lxm.xupt@gmail.com>

- [ ] Bugfix

___
### Bugfix
- Description
a lot of test cases with asset.Equal got wrong params with the expected and the actual value
especially with prs when test case run failed, the notice will be the wrong result
- How to fix?
change the params order


tao12345666333 (approved):
Thanks!

gxthrj (approved):

PR #827, opened 2022-01-03
feat: ensure that the lease can be actively released before program shutdown to reduce the time required for failover

nic-6443:
Ensure that the lease can be actively released before program shutdown to reduce the time required for failover

### Type of change:

- [x] Improve performance

### DIFF
--- a/pkg/ingress/controller.go
+++ b/pkg/ingress/controller.go
@@ -380,10 +380,7 @@ func (c *Controller) Run(stop chan struct{}) error {
 				c.apisix.DeleteCluster(c.cfg.APISIX.DefaultClusterName)
 			},
 		},
-		// Set it to false as current leaderelection implementation will report
-		// "Failed to release lock: resource name may not be empty" error when
-		// ReleaseOnCancel is true and the Run context is cancelled.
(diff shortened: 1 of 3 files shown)


pkg/ingress/controller.go
```diff
@@ -380,10 +380,7 @@ func (c *Controller) Run(stop chan struct{}) error {
 				c.apisix.DeleteCluster(c.cfg.APISIX.DefaultClusterName)
 			},
 		},
-		// Set it to false as current leaderelection implementation will report
-		// "Failed to release lock: resource name may not be empty" error when
-		// ReleaseOnCancel is true and the Run context is cancelled.
-		ReleaseOnCancel: false,
```
tokers, line 386: Is this error fixed?

nic-6443: It should be, I didn't find the error mentioned in the comment

tokers: OK.

tao12345666333 (approved):

starsz (approved):

tokers (approved):

Issue #883, opened 2022-02-22
bug: failed to list plugins' names


lingsamuel:
![ZQP6UFwF91](https://user-images.githubusercontent.com/14567045/155108623-3227c5bd-e06e-4408-af8e-45a400a285d7.png)

Using apache/apisix:dev and master branch apisix-ingress-controller.

tao12345666333:
Assigned. Thanks

PR #888, opened 2022-02-25
fix json unmarshal error when list plguins

cmssczy:
### Type of change:

- [x] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches

bug fix https://github.com/apache/apisix-ingress-controller/issues/883
the response body is JSON with unknown field names but not a string slice.
### Pre-submission checklist:

### DIFF
--- a/pkg/apisix/cluster.go
+++ b/pkg/apisix/cluster.go
@@ -750,10 +750,15 @@ func (c *cluster) getList(ctx context.Context, url, resource string) ([]string, 
 		return nil, err
 	}
 
-	var listResponse []string
+	var listResponse map[string]interface{}
 	dec := json.NewDecoder(resp.Body)
 	if err := dec.Decode(&listResponse); err != nil {
 		return nil, err
 	}
-	return listResponse, nil
+	res := make([]string, 0, len(listResponse))
+
+	for name := range listResponse {
+		res = append(res, name)
+	}
+	return res, nil
 }
--- a/pkg/apisix/plugin_test.go
+++ b/pkg/apisix/plugin_test.go
@@ -19,6 +19,7 @@ import (
 	"encoding/json"
 	"net/http"
 	"net/url"
+	"sort"
 	"strings"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


tao12345666333:
Thanks! It's on my list

pkg/apisix/cluster.go
```diff
@@ -750,10 +750,16 @@ func (c *cluster) getList(ctx context.Context, url, resource string) ([]string,
...
 	}
 
-	var listResponse []string
+	var listResponse map[string]interface{}
 	dec := json.NewDecoder(resp.Body)
 	if err := dec.Decode(&listResponse); err != nil {
 		return nil, err
 	}
-	return listResponse, nil
+	var res []string = make([]string, len(listResponse))
```
tokers, line 758:
```suggestion
	res []string := make([]string, 0, len(listResponse))
```

After that, just using `append`, we don't need the indice `i`.

tokers (commented):
@cmssczy Thanks for your PR, could you give us an example about the response? Fair to say, I didn't realize the difference.

cmssczy:
> could you give us an example about the response? Fair to say, I didn't realize the difference.

@tokers 
The response body json format is like:
```json
{
	"serverless-pre-function": {
		"version": 0.1,
		"priority": 10000
	},
	"echo": {
		"version": 0.1,
		"priority": 412
	}
}
```
It is a json struct with unknown field names but not a string slice. So I think we should use a `map` instead of `slice`.

pkg/apisix/cluster.go
cmssczy: Done

tao12345666333 (approved):
LGTM

lingsamuel:
We need a mechanism to ensure that the APISIX interface we are using is not outdated.

tao12345666333:
> We need a mechanism to ensure that the APISIX interface we are using is not outdated.

yep. we may need to convert directly from the upstream schema.

now let's move forward.

Issue #880, opened 2022-02-20
e2e: custom registry


lingsamuel:
Support for custom registries can make it easier to run e2e in existing clusters directly using ginkgo.

Belyenochi:
hi @lingsamuel , I tried adding custom registry support.

tao12345666333:
Thanks @Belyenochi

PR #896, opened 2022-03-02
feat: support custom registry for e2e test

Belyenochi:
### Type of change:

- [ ] Bugfix
- [x] New feature provided
- [ ] Improve performance
- [ ] Backport patches

Solved [#880](https://github.com/apache/apisix-ingress-controller/issues/880)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,7 +69,7 @@ e2e-test: ginkgo-check push-images
 	kubectl apply -k $(PWD)/samples/deploy/crd
 	cd test/e2e \
 		&& go mod download \
-		&& export CUSTOM_REGISTRY=$(CUSTOM_REGISTRY)
+		&& export REGISTRY=$(REGISTRY)
 		&& ACK_GINKGO_RC=true ginkgo -cover -coverprofile=coverage.txt -r --randomizeSuites --randomizeAllSpecs --trace --nodes=$(E2E_CONCURRENCY)
 
 .PHONY: ginkgo-check
--- a/Makefile
+++ b/Makefile
@@ -63,7 +63,7 @@ lint:
 unit-test:
 	go test -cover -coverprofile=coverage.txt ./...
 
-### e2e-test:             Run e2e test cases (kind is required)
+### e2e-test:             Run e2e test cases (in existing clusters directly)
 .PHONY: e2e-test
 e2e-test: ginkgo-check push-images
 	kubectl apply -k $(PWD)/samples/deploy/crd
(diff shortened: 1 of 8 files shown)


tao12345666333:
Thanks! it's on my list. 

I will review it ASAP.

Makefile
```diff
@@ -19,6 +19,7 @@ default: help
 VERSION ?= 1.4.0
 RELEASE_SRC = apache-apisix-ingress-controller-${VERSION}-src
 LOCAL_REGISTRY="localhost:5000"
+CUSTOM_REGISTRY ?= ""
```
lingsamuel, line 22: Instead of using a new environment variable, can we change the old one to a more generic name? The new variable does not seem necessary.

tao12345666333: +1

Belyenochi: Thanks for review @tao12345666333 @lingsamuel, I've tried renaming REGISTRY, here I'm worried if the kind-up part is needed, I think we should check if REGISTRY points to kind and then run the judgment?

lingsamuel: IMO, it would be better to move the kind-up from the prerequisite of push-images to the e2e-test. Since the users may indeed have a registry at localhost:5000.
Users can then run the e2e like this: build the images, export the REGISTRY env var, run `make push-images`, and then execute ginkgo directly.
Or we could provide a new entry such as `e2e-test-local` just like the e2e-test but without kind-up.

Belyenochi: > IMO, it would be better to move the kind-up from the prerequisite of push-images to the e2e-test. Since the users may indeed have a registry at localhost:5000. Users can then run the e2e like this: build the images, export the REGISTRY env var, run `make push-images`, and then execute ginkgo directly. Or we could provide a new entry such as `e2e-test-local` just like the e2e-test but without kind-up.

@lingsamuel Thanks for the suggestion, e2e-test-local has been re-added.

lingsamuel (approved):
LGTM
We don't need to change CI config since it starts kind directly https://github.com/apache/apisix-ingress-controller/blob/master/.github/workflows/e2e-test-ci.yml#L89

Makefile
```diff
@@ -63,52 +63,57 @@ lint:
...
 
-### e2e-test:             Run e2e test cases (kind is required)
+### e2e-test:             Run e2e test cases (in existing clusters directly)
 .PHONY: e2e-test
-e2e-test: ginkgo-check push-images-to-kind
+e2e-test: ginkgo-check push-images
 	kubectl apply -k $(PWD)/samples/deploy/crd
 	cd test/e2e \
 		&& go mod download \
+		&& export REGISTRY=$(REGISTRY)
```
tao12345666333, line 72:
```suggestion
		&& export REGISTRY=$(REGISTRY) \
```

need to be modified here.

tao12345666333 (changes requested):

Makefile
Belyenochi: Done

tao12345666333 (approved):

lingsamuel:
Thanks!

PR #910, opened 2022-03-09
fix: Fix response error rate expression on grafana

dgff07:
### Type of change:

- [x] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches

If we check the metrics available, we have:
* apisix_ingress_controller_apisix_bad_status_codes
* apisix_ingress_controller_apisix_requests

First, I think the `apisix_ingress_controller_apisix_bad_status_codes` is not a good name since it counts all the requests including the good ones.
The `apisix_ingress_controller_apisix_requests` seems to only have the number of successful requests.

Currently, the response error rate is calculated using the `apisix_ingress_controller_apisix_requests`, which I believe is not correct since it doesn't have the total requests made.

So, I changed the expression to use only the `apisix_ingress_controller_apisix_bad_status_codes` and put also the possibility to define a time range.

### Pre-submission checklist:

* [x] Have you modified the corresponding document?

### DIFF
--- a/docs/assets/other/json/apisix-ingress-controller-grafana.json
+++ b/docs/assets/other/json/apisix-ingress-controller-grafana.json
@@ -414,7 +414,7 @@
       "targets": [
         {
           "exemplar": true,
-          "expr": "sum(apisix_ingress_controller_apisix_bad_status_codes{status_code =~ \"[4-5].*\"}) / sum(apisix_ingress_controller_apisix_requests)",
+          "expr": "sum(increase(apisix_ingress_controller_apisix_bad_status_codes{status_code =~ \"[4-5].*\"}[$__range])) / sum(increase(apisix_ingress_controller_apisix_bad_status_codes[$__range]))",
           "hide": false,
           "interval": "",
           "legendFormat": "",


tao12345666333:
Thanks, let me check.

Sindweller:
Agreed with
> First, I think the apisix_ingress_controller_apisix_bad_status_codes is not a good name since it counts all the requests including the good ones.

And this has been modified in #1012 

---

But IMOP `apisix_ingress_controller_apisix_requests` contains both success reqeusts and failed request. Here is a code slice of the exact increase of this count. 

```
if err := u.cluster.deleteResource(ctx, url, "upstream"); err != nil {
		u.cluster.metricsCollector.IncrAPISIXRequest("upstream")
		return err
	}
u.cluster.metricsCollector.IncrAPISIXRequest("upstream")
return nil
```

and in `deleteResource` we consider status code not in [200, 204, 404] as err:

```
if resp.StatusCode != http.StatusOK && resp.StatusCode != http.StatusNoContent && resp.StatusCode != http.StatusNotFound {
		message := readBody(resp.Body, url)
		if c.isFunctionDisabled(message) {
			return ErrFunctionDisabled
		}
		err = multierr.Append(err, fmt.Errorf("unexpected status code %d", resp.StatusCode))
		err = multierr.Append(err, fmt.Errorf("error message: %s", message))
		if strings.Contains(message, "still using") {
			return cache.ErrStillInUse
		}
		return err
	}
```

Either success or failed request was send, this metric would increase.

So the "expr" now: `"sum(apisix_ingress_controller_apisix_bad_status_codes{status_code =~ \"[4-5].*\"}) / sum(apisix_ingress_controller_apisix_requests)"` is correct, I think.

PR #949, opened 2022-04-05
Split e2e test cases

fgksgf:
### Type of change:

Split e2e test cases to reduce overall time consumption.

### What this PR does / why we need it:

Close #932 

### Pre-submission checklist:

* [x] Did you explain what problem does this PR solve? Or what new features have been added?
* [x] Have you added corresponding test cases?
* [x] Have you modified the corresponding document?
* [x] Is this PR backward compatible? **If it is not backward compatible, please discuss on the [mailing list](https://github.com/apache/apisix-ingress-controller#community) first**

PR #957, opened 2022-04-08
Add a pre-check for E2E tests

fgksgf:
### Type of change:

- [x] CI

In #949 we use `ginkgo --focus` to run test cases in parallel, but some newly added cases may not follow the naming convention. So we need to add a pre-check to avoid some E2E cases not running.

### Pre-submission checklist:

### Test

Tested on Linux:
image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -205,7 +205,7 @@ update-gofmt:
 .PHONY: update-all
 update-all: update-codegen update-license update-mdlint update-gofmt
 
-### names-check:          Check if e2e test cases' names have the prefix "suite-<suite-name>".
-.PHONY: names-check
-names-check:
+### e2e-names-check:          Check if e2e test cases' names have the prefix "suite-<suite-name>".
+.PHONY: e2e-names-check
+e2e-names-check:
 	chmod +x ./utils/check-e2e-names.sh && ./utils/check-e2e-names.sh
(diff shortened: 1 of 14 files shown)


fgksgf:
@tao12345666333 plz add a reviewer so that the e2e CI job can run.

fgksgf:
The script can find we missed the prefix in some files:
image

tao12345666333:
> The script can find we missed the prefix in some files

Great!!

tao12345666333:
re-run CI.

tao12345666333 (commented):
Should we add a `check-e2e-name` rule to the Makefile? It might be more convenient

fgksgf:
> Should we add a `check-e2e-name` rule to the Makefile? It might be more convenient

Good idea

tao12345666333 (approved):
LGTM

Makefile
```diff
@@ -203,3 +204,8 @@ update-gofmt:
 ### update-all:           Update all update- rules.
 .PHONY: update-all
 update-all: update-codegen update-license update-mdlint update-gofmt
+
+### names-check:          Check if e2e test cases' names have the prefix "suite-<suite-name>".
+.PHONY: names-check
```
lingsamuel, line 209: I think `e2e-names-check` is better because it explicitly indicates that it is related to the e2e test.

lingsamuel (approved):

Makefile
fgksgf: ok, i'll fix it

PR #1006, opened 2022-05-08
feat: Add support force-ssl-redirect to ingress translator

jaysonsantos:
### Type of change:

- [ ] Bugfix
- [x] New feature provided
- [ ] Improve performance
- [ ] Backport patches

This should make apsix use the annotation ingress.kubernetes.io/force-ssl-redirect: true
and convert it to a redirect plugins, making it compatible with common annotations for ingresses.

### Pre-submission checklist:

* [x] Have you added corresponding test cases?
* [x] Have you modified the corresponding document?
* [x] Is this PR backward compatible? **If it is not backward compatible, please discuss on the [mailing 
list](https://github.com/apache/apisix-ingress-controller#community) first**

### DIFF
--- a/docs/en/latest/concepts/annotations.md
+++ b/docs/en/latest/concepts/annotations.md
@@ -104,6 +104,7 @@ Redirect
 You can use the following annotations to control the redirect behavior.
 
 * `k8s.apisix.apache.org/http-to-https`
+* `ingress.kubernetes.io/force-ssl-redirect`
 
 If this annotation set to `true` and the request is HTTP, it will be automatically redirected to HTTPS with 301 response code,
 and the URI will keep the same as client request.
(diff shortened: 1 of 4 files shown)


tao12345666333:
Thanks for your contribution.

I understand that you want to make APISIX Ingress compatible with Ingress NGINX better, right?

As you may have noticed, the annotation used in APISIX Ingress is completely different from Ingress NGINX (especially its prefix)

This is a deliberate distinction. 

Currently we have no plans to make annotations fully compatible, but **I think we can discuss this and see if it has value, and some practical scenarios or use cases**

jaysonsantos:
Hey there, that is right, this would be something to try and ease out on the migration of existing services.
But if the idea is to have its own namespace, i guess there is no value in this pr.
Thanks for the answer!

tao12345666333:
Thanks!
Maybe we can provide a tool to complete the conversion of Ingress + annotation to ApisixRoute configuration?

Issue #1041, opened 2022-05-24
CI: Upgrade ginkgo to v2


tao12345666333:
Now we use ginkgo v1.16.4 version in e2e,

https://github.com/apache/apisix-ingress-controller/blob/f6f0a3b5552ba8fda556e0edb9296c1c0a4c3e31/test/e2e/go.mod#L11

v1 has been formally deprecated. So we need to Upgrade ginkgo to v2

Here is an upgrade guide for reference

https://onsi.github.io/ginkgo/MIGRATING_TO_V2

tao12345666333:
#1046 has been merged, I will close this one. Thanks @Fatpa

PR #1046, opened 2022-05-26
e2e-test: upgrade to ginkgo v2

Fatpa:
### Type of change:

- [ ] Bugfix
- [x] New feature provided
- [ ] Improve performance
- [ ] Backport patches

[#1041 ](https://github.com/apache/apisix-ingress-controller/issues/1041)

### Pre-submission checklist:


tao12345666333:
re-run all jobs

test/e2e/go.mod
```diff
@@ -8,7 +8,7 @@ require (
 	github.com/gavv/httpexpect/v2 v2.2.0
 	github.com/gorilla/websocket v1.4.2
 	github.com/gruntwork-io/terratest v0.32.8
-	github.com/onsi/ginkgo v1.16.4
+	github.com/onsi/ginkgo/v2 v2.1.4
```
tokers, line 11: Can we use `replace` directive so that we can avoid a lot of changes?

Fatpa: I don't think `replace` is the best practice for the gomod.
And if we use `replace`, it may lead to some confusion in the source code.

Fatpa: I use an alias on the import, so it has already avoid as more changes as we can.

test/e2e/go.sum
```diff
@@ -542,13 +543,18 @@ github.com/onsi/ginkgo v1.12.1/go.mod h1:zj2OWP4+oCPe1qIXoGWkgMRwljMUYCdkwsT2108
 github.com/onsi/ginkgo v1.14.0/go.mod h1:iSB4RoI2tjJc9BBv4NKIKWKya62Rps+oPG/Lv9klQyY=
```
tao12345666333, line 543: we cloud clean older version, right?

tao12345666333 (commented):
Please merge master, we have fixed the test case #1055

test/e2e/go.mod
tao12345666333: I agree with @Fatpa 
I hope this project can be more clear, and there are not many changes, which have been done in this PR.

test/e2e/go.sum
Fatpa: We can't. 
I try to remove it from the `go.sum`, but after re-running `go mod tidy`, it will come back again.

Fatpa:
Still has some unknown failure checks. Maybe due to the resource lack of the e2e test environment?

PR #1149, opened 2022-07-13
ci: upgrade e2e-test-ci

AlinsRan:
### Type of change:

- [x] Bugfix
- [ ] New feature provided
- [x] Improve performance
- [ ] Backport patches

ref: #1164 

* Add retry mechanism to reduce random errors.
* Fix the wrong operation of e2e using `EnsureNum...` func to ensure resource update.
* Ci reuses the built image, which has reduced the build time of the job.

### It will retry 6 times to ensure the stability above 80%

 - [x] 1th - pass
 - [x] 2th - pass
 - [x] 3th - pass
 - [x] 4th - pass
 - [x] 5th - pass
 - [x] done

retry 6 times:
pass: 6
faild: 0


lingsamuel (approved):

Makefile
```diff
@@ -49,11 +50,67 @@ build:
...
+	docker tag test-backend:$(IMAGE_TAG) $(REGISTRY)/test-backend:$(IMAGE_TAG)
+
+	docker build -t apache/apisix-ingress-controller:$(IMAGE_TAG) --build-arg ENABLE_PROXY=$(ENABLE_PROXY) .
+	docker tag apache/apisix-ingress-controller:$(IMAGE_TAG) $(REGISTRY)/apache/apisix-ingress-controller:$(IMAGE_TAG)
+
+	docker pull jmalloc/echo-server:latest
+	docker tag  jmalloc/echo-server:latest $(REGISTRY)/jmalloc/echo-server:$(IMAGE_TAG)
+
+	docker pull busybox:1.28
+	docker tag  busybox:1.28 $(REGISTRY)/busybox:$(IMAGE_TAG)
```
tao12345666333, line 98: I do not recommend unifying the tags of these images into dev.
If the task fails, we need to check the code to know which image is used.

Can you explain your benefit of unifying them?

tao12345666333 (commented):
I'm guessing it's because there are places in the code that need to be marked with their image tags?

Makefile
AlinsRan: The advantage of this is that because we use a local repository, we don't need to care about these images tags in the code. When upgrading, we only need to change the `Makefile` file. 
At present, `localhost:5000` is used to identify the local warehouse in our code. It's not difficult to track this problem.

tao12345666333: make sense, let's move forward.

tao12345666333 (approved):

PR #1152, opened 2022-07-14
fix: trigger ApisixRoute event when service is created

### LINKED ISSUE #926: bug: we should create APISIX routes even if Service is created later
> https://github.com/apache/apisix-ingress-controller/blob/c9e0c965cd2226e2c44aa4e692a8bf57d1586aa3/pkg/ingress/apisix_route.go#L261-L271
> 
> We will delete this event from workqueue if we have an `ApisixRoute` resources without any valid backends. It is reasonable?

@cmssczy  This change is to fix the #810 and was introduced in #893 

**This does cause some problems.**

For example, create an ApisixRoute first, which references a service that does not exist.
But when a Service is created, the APISIX Ingress controller can't actually create a route.

_Originally posted by @tao12345666333 in https://github.com/apache/apisix-ingress-controller/issues/925#issuecomment-1072128249_

lingsamuel:
### Type of change:

Fixes #926

- [x] Bugfix


test/e2e/suite-ingress/suite-ingress-resource/resourcepushing.go
```diff
@@ -576,6 +578,115 @@ spec:
...
+`, "httpbin-temp", 80)
+
+			assert.Nil(ginkgo.GinkgoT(), s.CreateVersionedApisixResource(apisixRoute), "creating ApisixRoute")
+			// We don't have service yet, so upstream==0
+			err := s.EnsureNumApisixRoutesCreated(0)
+			assert.Nil(ginkgo.GinkgoT(), err, "Checking number of routes")
+			err = s.EnsureNumApisixUpstreamsCreated(0)
+			assert.Nil(ginkgo.GinkgoT(), err, "Checking number of upstreams")
+			err = s.EnsureNumApisixPluginConfigCreated(0)
+			assert.Nil(ginkgo.GinkgoT(), err, "Checking number of pluginConfigs")
```
tao12345666333, line 679: Actually we don't create PluginConfig, so this can be deleted

pkg/ingress/apisix_route.go
```diff
@@ -41,13 +43,20 @@ type apisixRouteController struct {
 	controller *Controller
 	workqueue  workqueue.RateLimitingInterface
 	workers    int
+
+	// svc meta key -> ApisixRoute name -> empty
+	// map[string]map[string]struct{}
+	svcLock sync.RWMutex
+	svcMap  map[string]map[string]struct{} //*sync.Map
```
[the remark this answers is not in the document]
lingsamuel: I would like to use more specific types instead of the untyped sync.Map.

pkg/kube/apisix_route.go
```diff
@@ -35,6 +35,12 @@ type ApisixRouteLister interface {
 	V2beta3(string, string) (ApisixRoute, error)
 	// V2 gets the ApisixRoute in apisix.apache.org/v2.
 	V2(string, string) (ApisixRoute, error)
+	// V2beta2Lister gets the v2beta2 lister
+	V2beta2Lister() listersv2beta2.ApisixRouteLister
```
tao12345666333, line 39: we don't need this?

pkg/kube/apisix_route.go
```diff
@@ -78,6 +84,18 @@ type apisixRoute struct {
 	v2           *configv2.ApisixRoute
 }
 
+func (l *apisixRouteLister) V2beta2Lister() listersv2beta2.ApisixRouteLister {
+	return l.v2beta2Lister
+}
```
tao12345666333, line 89: ditto

test/e2e/suite-ingress/suite-ingress-resource/resourcepushing.go
```diff
@@ -576,6 +578,113 @@ spec:
...
+            timeoutSeconds: 2
+          readinessProbe:
+            failureThreshold: 3
+            initialDelaySeconds: 2
+            periodSeconds: 5
+            successThreshold: 1
+            tcpSocket:
+              port: 80
+            timeoutSeconds: 2
+          image: "localhost:5000/kennethreitz/httpbin"
```
AlinsRan, line 622: localhost:5000/kennethreitz/httpbin:dev

pkg/ingress/apisix_route.go
```diff
@@ -82,9 +95,114 @@ func (c *apisixRouteController) runWorker(ctx context.Context) {
...
+			newObj = ar.V2()
+		}
+
+		// calculate diff, so we don't need to care about the event order
+		if old != nil {
+			for _, rule := range old.Spec.HTTP {
+				for _, backend := range rule.Backends {
+					oldBackends = append(oldBackends, backend.ServiceName)
+
+					delete(c.svcMap[old.Namespace+"/"+backend.ServiceName], old.Name)
```
AlinsRan, line 175: Does this conflict with line 198? There is no lock here.

pkg/ingress/apisix_route.go
```diff
@@ -82,9 +95,114 @@ func (c *apisixRouteController) runWorker(ctx context.Context) {
...
+			newObj = ar.V2beta3()
+		}
+
+		// calculate diff, so we don't need to care about the event order
+		if old != nil {
+			for _, rule := range old.Spec.HTTP {
+				for _, backend := range rule.Backends {
+					oldBackends = append(oldBackends, backend.ServiceName)
+
+					delete(c.svcMap[old.Namespace+"/"+backend.ServiceName], old.Name)
```
AlinsRan, line 142: Does this conflict with line 198? There is no lock here.

pkg/ingress/apisix_route.go
```diff
@@ -41,13 +42,19 @@ type apisixRouteController struct {
 	controller *Controller
 	workqueue  workqueue.RateLimitingInterface
 	workers    int
+
+	// svc meta key -> ApisixRoute name -> empty
+	svcLock sync.RWMutex
```
AlinsRan, line 47: No read lock。

pkg/ingress/apisix_route.go
```diff
@@ -41,13 +42,19 @@ type apisixRouteController struct {
 	controller *Controller
 	workqueue  workqueue.RateLimitingInterface
 	workers    int
+
+	// svc meta key -> ApisixRoute name -> empty
```
AlinsRan, line 46: we should only maintain the `map` when the service is not exist? @tao12345666333

test/e2e/suite-ingress/suite-ingress-resource/resourcepushing.go
lingsamuel: This image doesn't have such tag

pkg/ingress/apisix_route.go
tao12345666333: no, the service can be modified.

Consider this one, when service be deleted and re-created, we also need the map

AlinsRan: OK.

tao12345666333 (approved):
LGTM

let's move forward.

gxthrj (approved):

Issue #1058, opened 2022-05-31
proposal: restructure `pkg/ingress`


lingsamuel:
As support for the Gateway API progressed, we needed to redesign the structure of `pkg/ingress`.
This directory now has multiple types of resources mixed together: Kubernetes native resources, APISIX Ingress resources, and Gateway API resources. More types of resources may be added in the future.
In addition, the `controller.go` file has become increasingly large.

We need to redesign the controller so that supporting different resources doesn't require massive changes to the outer files, and to keep the changes in a smaller module as much as possible.

tao12345666333:
+1
I agree with you.

Do you have a more specific plan? 
We can put this proposal into this year's roadmap

lingsamuel:
Perhaps we could create an interface called "(Feature)Provider". The provider holds several small controllers and associated clientet/listers/informers.

Also, we should separate the translators and only make the Kubernetes translator publicly accessible, since any other translator could potentially rely on it.
APISIX Ingress and any other CRD translators should be private.


- RootController
  - KubernetesProvider (public)
    - Controllers: secret, service, pod, endpoint...
    - clientset/listers/informers (public)
    - KubernetesTranslator (public)
  - ApisixIngressProvider
    - Controllers: APISIX Ingress CRDs
    - clientset/listers/informers
    - ApisixIngressTranslator
  - GatewayProvider
    - Controllers: Gateway CRDs
    - clientset/listers/informers
    - GatewayTranslator

tao12345666333:
SGTM

I prefer to rename KubernetesProvider to CommonProvider.

When dealing with various resources(Ingress/CR/Gateway resource), we may need to deal with these basic resources of Kubernetes.

PR #1204, opened 2022-08-02
Restruct pkg/ingress

lingsamuel:
### Type of change:

- [x] Refactor

Part of #1058 

*Submit PR for CI, I will update the changes later*

This is a huge PR, so I will describe the changes and ideas in detail.

- This PR is expected to be merged in 1.16, so I delete the `apisix-route-version` option. If this option should be kept for more versions, I will restore it.

- This PR splits `pkg/ingress` into multiple providers, including `apisix`, `ingress` and `k8s`, and `gateway-api` was also split in a previous PR.
  - Controllers are no longer dependent on the parent controller. If a controller needs data held by other controllers, the data should be exposed by their provider interface. (See NamespaceProvider and PodProvider)
  - There are separate translators for "apisix" and "ingress". Since this PR doesn't change any detailed translation logic, ingress translator depends on apisix translator for now.
  - The CRD controllers handle own status records instead of using the shared type-switch function.
  - Some shared variables, such as MetricsCollector, KubeClient, and config, are embedded by commonConfig.

- If the E2E_ENV is not `ci`, replace ingress-controller deployment yaml imagePullPolicy to `Always`.

### Pre-submission checklist:


AlinsRan (approved):

pkg/providers/apisix/provider.go
```diff
@@ -0,0 +1,202 @@
...
+
+var _ Provider = (*apisixProvider)(nil)
+
+type Provider interface {
+	providertypes.Provider
+
+	Init(ctx context.Context) error
+	ResourceSync()
+
+	GetSslFromSecretKey(string) *sync.Map
```
tao12345666333, line 55: just a question, why do we put it here, is it a generic method?

tao12345666333 (approved):
LGTM

let's move forward

gxthrj (approved):

Issue #1201, opened 2022-08-01
test: Discuss how to face the docker compose update


Chever-John:
### Issue description

When I was preparing to build the development environment on my new host, I was surprised to find a simple ci build problem after running the make-e2e-test-local command.

This is easy to fix, the official command seems to have been updated from `docker-compose` to `docker compose`.

I can install an older version of Docker Compose, or just modify the makefile to fix it. But for future compatibility. I suggest to think about rewriting part of the logic on this side?

So do we need to have a discussion about this situation?

### Environment

- your apisix-ingress-controller version (output of apisix-ingress-controller version --long):
- your Kubernetes cluster version (output of kubectl version):
- if you run apisix-ingress-controller in Bare-metal environment, also show your OS version (uname -a):

### Minimal test code / Steps to reproduce

1. run the command `make e2e-test-local`;

### Actual result

```bash
./test/e2e/testdata/wolf-rbac/cmd.sh: line 33: docker-compose: command not found
./test/e2e/testdata/wolf-rbac/cmd.sh: line 36: docker-compose: command not found
curl: (7) Failed connect to 127.0.0.1:12180; Connection refused
curl: (7) Failed connect to 127.0.0.1:12180; Connection refused
curl: (7) Failed connect to 127.0.0.1:12180; Connection refused
make: *** [e2e-wolf-rbac] Error 7
```

### Error log

none

### Expected result

Success!

AlinsRan:
You have not installed docker-compose.

Chever-John:
@AlinsRan 
Maybe I didn't express myself clearly. What I meant was that the latest version of docker compose executes the command `docker compose` instead of `docker-compose`. So do we have to improve the makefile to accommodate this change. Or does anyone have a better suggestion?

zaunist:
> First, I have installed Docker Compose!
> 
> @AlinsRan Maybe I didn't express myself clearly. What I meant was that the latest version of docker compose executes the command `docker compose` instead of `docker-compose`. So do we have to improve the makefile to accommodate this change. Or does anyone have a better suggestion?

Hello, @Chever-John. 

I went through the [Docker documentation](https://docs.docker.com/compose/) and noticed the following passage

> Run docker compose up and the [Docker compose command](https://docs.docker.com/compose/#compose-v2-and-the-new-docker-compose-command) starts and runs your entire app. You can alternatively run docker-compose up using Compose standalone(docker-compose binary).

[...]

PR #1213, opened 2022-08-04
feat: support docker-compose command detect

zaunist:
### Type of change:

- [ ] Bugfix
- [x] New feature provided
- [ ] Improve performance
- [ ] Backport patches

fix https://github.com/apache/apisix-ingress-controller/issues/1201

### Pre-submission checklist:

### DIFF
--- a/test/e2e/testdata/wolf-rbac/cmd.sh
+++ b/test/e2e/testdata/wolf-rbac/cmd.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 #
 # Licensed to the Apache Software Foundation (ASF) under one or more
--- a/test/e2e/testdata/wolf-rbac/cmd.sh
+++ b/test/e2e/testdata/wolf-rbac/cmd.sh
@@ -20,26 +20,26 @@
 cd test/e2e/testdata/wolf-rbac/
 
 OPTION=$1
-COMPOSE="docker-compose"
+COMPOSE=(docker-compose)
 
 if  ! command -v "docker-compose" > /dev/null; then
 	echo "docker-compose not found, try docker compose"
-    COMPOSE="docker compose"
+    COMPOSE=(docker-compose)
 fi
 
 if  [ $OPTION = "ip" ]; then
     echo -n `docker inspect -f '{{range .NetworkSettings.Networks}}{{.Gateway}}{{end}}' wolf-server`
 elif [ $OPTION = "start" ]; then
-    eval "$COMPOSE -f 'docker-compose.yaml'  -p 'wolf-rbac' down"
+    $COMPOSE -f 'docker-compose.yaml'  -p 'wolf-rbac' down
     rm -rf db-psql.sql
 
     wget https://raw.githubusercontent.com/iGeeky/wolf/f6ddeb75a37bff90406f0f0a2b7ae5d16f6f3bd4/server/script/db-psql.sql
 
     # start database
-    eval "$COMPOSE up -d database"
+    $COMPOSE up -d database
 
     # start wolf-server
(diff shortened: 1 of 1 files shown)


test/e2e/testdata/wolf-rbac/cmd.sh
```diff
@@ -21,6 +21,11 @@ cd test/e2e/testdata/wolf-rbac/
 
 OPTION=$1
 
+if  ! command -v "docker-compose" > /dev/null; then
+	echo "docker-compose not found, please install docker-compose"
```
Chever-John, line 25: I don't think it should be resolved that way. We should write more judgments, because my current situation is that I only have the `docker compose` command, not `docker-compose`, and I just need to replace `docker-compose` with `docker compose`.

Chever-John (commented):
I think this is a wrong solution.

test/e2e/testdata/wolf-rbac/cmd.sh
zaunist: If so, we need to write both docker compose and docker-compose logic, which would be too cumbersome, and I noticed that docker compose is only integrated in docker desktop for Windows and Mac OS, not for Linux systems. 

 Ref https://github.com/docker/compose#windows-and-macos

Chever-John: It is not necessary to ah, define a global variable. If `docker-compose` exists, copy `docker-compose` to this variable. If not, continue to determine if `docker compose` exists, and if it does, assign `docker compose` to this variable. 

Finally, replace the original `docker-compose` globally.

zaunist: > 

agree

test/e2e/testdata/wolf-rbac/cmd.sh
```diff
@@ -20,20 +20,26 @@
...
+if  ! command -v "docker-compose" > /dev/null; then
+	echo "docker-compose not found, try docker compose"
+    COMPOSE="docker compose"
+fi
 
 if  [ $OPTION = "ip" ]; then
     echo -n `docker inspect -f '{{range .NetworkSettings.Networks}}{{.Gateway}}{{end}}' wolf-server`
 elif [ $OPTION = "start" ]; then
-    docker-compose -f 'docker-compose.yaml'  -p 'wolf-rbac' down
+    eval "$COMPOSE -f 'docker-compose.yaml'  -p 'wolf-rbac' down"
```
tao12345666333, line 33: We don't need "eval"

zaunist: Due to  the `compose` is a subcommand of `docker`, a form like this would be wrong, 

```
cmd='docker compose'
$cmd xxx
```

e.g.

```
~                                                                                                                     ⍉
▶ cm='docker compose'

~
▶ $cm version
zsh: command not found: docker compose

~                                                                                                                     ⍉
▶ eval $cm version
Docker Compose version v2.7.0

```

zaunist: If we don't use eval, is there a better way to do it?

zaunist: @Chever-John What do you think?

lingsamuel:
```bash
cmd=( docker compose )
```

zaunist: > ```shell
> cmd=( docker compose )
> 
> $cmd
> ```

This method doesn't work. I can't think of any other way than eval.

image

zaunist:
I don't have time to update the code

PR #1311, opened 2022-09-01
fix: ingress class not effect in resource sync logic

### LINKED ISSUE #1310: bug: ingress filter is not handled in resource sync logic
### Issue description

Apisix ingress controller can filter other ingress controller object by setting ingressclassname. But it seems not work in 1.5.0-rc1. I checked the code and found the filter is implement by 

```
func (c *ingressController) isIngressEffective(ing kube.Ingress) bool {
	var (
		ic  *string
		ica string
	)
	if ing.GroupVersion() == kube.IngressV1 {
		ic = ing.V1().Spec.IngressClassName
		ica = ing.V1().GetAnnotations()[_ingressKey]
	} else if ing.GroupVersion() == kube.IngressV1beta1 {
		ic = ing.V1beta1().Spec.IngressClassName
		ica = ing.V1beta1().GetAnnotations()[_ingressKey]
	} else {
		ic = ing.ExtensionsV1beta1().Spec.IngressClassName
		ica = ing.ExtensionsV1beta1().GetAnnotations()[_ingressKey]
	}

	// kubernetes.io/ingress.class takes the precedence.
	if ica != "" {
		return ica == c.controller.cfg.Kubernetes.IngressClass
	}
	if ic != nil {
		return *ic == c.controller.cfg.Kubernetes.IngressClass
	}
	return false
}
```

and called in onadd, ondelete, onupdate

but in the new func 

```
func (c *ingressController) ResourceSync() {
	objs := c.controller.ingressInformer.GetIndexer().List()
	for _, obj := range objs {
		key, err := cache.MetaNamespaceKeyFunc(obj)
		if err != nil {
			log.Errorw("found Ingress resource with bad meta namespace key", zap.String("error", err.Error()))
			continue
		}
		if !c.controller.isWatchingNamespace(key) {
			continue
		}
		ing := kube.MustNewIngress(obj)
		c.workqueue.Add(&types.Event{
			Type: types.EventAdd,
			Object: kube.IngressEvent{
				Key:          key,
				GroupVersion: ing.GroupVersion(),
			},
		})
	}
}
```

it did't do such check, is this a bug?

### Environment

- your apisix-ingress-controller version (output of apisix-ingress-controller version --long):
- your Kubernetes cluster version (output of kubectl version):
- if you run apisix-ingress-controller in Bare-metal environment, also show your OS version (uname -a):

xiangtianyu:
### Type of change:

- [x] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches

fix #1310 
### Pre-submission checklist:

### DIFF
--- a/pkg/providers/ingress/ingress.go
+++ b/pkg/providers/ingress/ingress.go
@@ -429,15 +429,12 @@ func (c *ingressController) ResourceSync() {
 			continue
 		}
 		ing := kube.MustNewIngress(obj)
-		valid := c.isIngressEffective(ing)
-		if valid {
-			log.Debugw("ingress add event arrived",
-				zap.Any("object", obj),
-			)
-		} else {
-			// resource sync do not care about the ingress class mismatch
+		if !c.isIngressEffective(ing) {
 			return
 		}
+		log.Debugw("ingress add event arrived",
+			zap.Any("object", obj),
+		)
 		c.workqueue.Add(&types.Event{
 			Type: types.EventAdd,
 			Object: kube.IngressEvent{
--- a/pkg/providers/ingress/ingress.go
+++ b/pkg/providers/ingress/ingress.go
@@ -286,12 +286,7 @@ func (c *ingressController) onAdd(obj interface{}) {
 	}
 
 	ing := kube.MustNewIngress(obj)
-	valid := c.isIngressEffective(ing)
-	if valid {
-		log.Debugw("ingress add event arrived",
-			zap.Any("object", obj),
-		)
-	} else {
+	if !c.isIngressEffective(ing) {
 		log.Debugw("ignore noneffective ingress add event",
 			zap.Any("object", obj),
 		)
@@ -324,13 +319,7 @@ func (c *ingressController) onUpdate(oldObj, newObj interface{}) {
 	if !c.namespaceProvider.IsWatchingNamespace(key) {
 		return
 	}
-	valid := c.isIngressEffective(curr)
-	if valid {
-		log.Debugw("ingress update event arrived",
-			zap.Any("new object", newObj),
-			zap.Any("old object", oldObj),
-		)
-	} else {
+	if !c.isIngressEffective(curr) {
 		log.Debugw("ignore noneffective ingress update event",
 			zap.Any("new object", oldObj),
 			zap.Any("old object", newObj),
@@ -368,12 +357,7 @@ func (c *ingressController) OnDelete(obj interface{}) {
 	if !c.namespaceProvider.IsWatchingNamespace(key) {
 		return
 	}
-	valid := c.isIngressEffective(ing)
-	if valid {
-		log.Debugw("ingress delete event arrived",
-			zap.Any("final state", ing),
-		)
-	} else {
+	if !c.isIngressEffective(ing) {
 		log.Debugw("ignore noneffective ingress delete event",
 			zap.Any("object", ing),
 		)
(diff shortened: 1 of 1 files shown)


pkg/providers/ingress/ingress.go
```diff
@@ -428,7 +428,20 @@ func (c *ingressController) ResourceSync() {
...
+		if valid {
+			log.Debugw("ingress add event arrived",
+				zap.Any("object", obj),
+			)
+		} else {
+			log.Debugw("ignore noneffective ingress add event",
+				zap.Any("object", obj),
+			)
+			return
+		}
```
tao12345666333, line 443: We can simplify the code here, right?

tao12345666333 (commented):
thanks for your contribution!

pkg/providers/ingress/ingress.go
xiangtianyu: Are you means to summarize these "valid judge" logic to common codes?

tao12345666333: yes, we don't need judge `valid` and print log.

xiangtianyu: > yes, we don't need judge `valid=true` and print log.

Ok, i'll modify it

pkg/providers/ingress/ingress.go
```diff
@@ -286,12 +286,7 @@ func (c *ingressController) onAdd(obj interface{}) {
 	}
 
 	ing := kube.MustNewIngress(obj)
-	valid := c.isIngressEffective(ing)
-	if valid {
-		log.Debugw("ingress add event arrived",
-			zap.Any("object", obj),
-		)
-	} else {
+	if !c.isIngressEffective(ing) {
```
AlinsRan, line 289: These places do not need to be changed, and the debug log needs to be kept.

pkg/providers/ingress/ingress.go
```diff
@@ -324,13 +319,7 @@ func (c *ingressController) onUpdate(oldObj, newObj interface{}) {
...
 		return
 	}
-	valid := c.isIngressEffective(curr)
-	if valid {
-		log.Debugw("ingress update event arrived",
-			zap.Any("new object", newObj),
-			zap.Any("old object", oldObj),
-		)
-	} else {
+	if !c.isIngressEffective(curr) {
```
AlinsRan, line 322: Ditto.

pkg/providers/ingress/ingress.go
```diff
@@ -428,7 +412,15 @@ func (c *ingressController) ResourceSync() {
 		if !c.namespaceProvider.IsWatchingNamespace(key) {
 			continue
 		}
+
 		ing := kube.MustNewIngress(obj)
+		if !c.isIngressEffective(ing) {
+			log.Debugw("ignore noneffective ingress add event",
```
AlinsRan, line 418: Only the log of the sync object should be printed, and the sync should not care about the ingresssclass log.

tao12345666333:
I agree with @AlinsRan , in this PR you can just keep the changes in the ResourceSync function, which will help me cherry-pick it to other versions.

pkg/providers/ingress/ingress.go
xiangtianyu: I misunderstanding sonething. And i will modify it soon

lingsamuel (approved):

pkg/providers/ingress/ingress.go
```diff
@@ -429,6 +429,15 @@ func (c *ingressController) ResourceSync() {
...
 		ing := kube.MustNewIngress(obj)
+		valid := c.isIngressEffective(ing)
+		if valid {
+			log.Debugw("ingress add event arrived",
+				zap.Any("object", obj),
+			)
+		} else {
+			// resource sync do not care about the ingress class mismatch
+			return
+		}
```
AlinsRan, line 440:
```suggestion
		if !c.isIngressEffective(ing) {
			return
		}
		log.Debugw("ingress add event arrived",
			zap.Any("object", obj),
		)
```

AlinsRan (approved):

tao12345666333 (approved):
LGTM

PR #1399, opened 2022-10-21
docs: update annotations page

navendu-pottekkat:
Signed-off-by: Navendu Pottekkat <navendu@apache.org>

Updates "Annotations" page.

Moves the page out of the "Concepts" section as a separate reference guide.


docs/en/latest/config.json
```diff
@@ -60,10 +60,13 @@
...
         "concepts/apisix_tls",
-        "concepts/apisix_cluster_config",
-        "concepts/annotations"
+        "concepts/apisix_cluster_config"
       ]
     },
+    {
+      "type": "doc",
+      "id": "annotations"
+    },
```
tao12345666333, line 69: thanks for your contribution

But I think annotations are more appropriate in concepts.
Is there any reason for you to separate it into a separate page?

navendu-pottekkat: Annotations are used in Kubernetes native Ingress resource while rest of the concepts is specific to APISIX's CRDs. IMO, annotations are a different entity. And it does not belong together with the concepts. I'm just thinking this out loud. Would like to hear your thoughts on this as well.

tao12345666333: Although annotations are a concept in Kubernetes, the configuration items that users can use are specific to APISIX Ingress. 
Users need learn these "concepts" when using this project, 
and we don't need to introduce the Ingress resource, because it is an existing concept in Kubernetes, and there is nothing special in the APISIX Ingress project,
what do you think?

navendu-pottekkat: Okay. That kinda makes sense. I will move the page back.

navendu-pottekkat:
@tao12345666333 I have moved back the page to under concepts.

docs/en/latest/config.json
tao12345666333: thanks

tao12345666333 (approved):

juzhiyuan (approved):
I think it's good to go, the failed lint is because of another md file.

navendu-pottekkat:
@tao12345666333 Now that I think about it, one of the reasons I thought moving the annotations page to a separate root-level page was discoverability for the user. An APISIX Ingress user might want to look up the annotations available and it might not be intuitive to check under the concepts section.

PR #1445, opened 2022-11-09
chore: gateway-api v0.5.1

AlinsRan:
### Type of change:

- [ ] Bugfix
- [ ] New feature provided
- [x] Improve performance
- [ ] Backport patches

### Pre-submission checklist:


Makefile
```diff
@@ -258,3 +266,24 @@ kind-load-images:
...
+
+### install:				Install Gateway API CRDs into the K8s cluster.
+.PHONY: install-gateway-api
+install-gateway-api: go-mod-download-gateway-api
+	kubectl apply -k $(GATEWAY_API_CRDS_LOCAL_PATH)
+	kubectl apply -k $(GATEWAY_API_CRDS_LOCAL_PATH)/experimental
+
+### uninstall-gateway-api:	Uninstall Gateway API CRDs from the K8s cluster.
+.PHONY: uninstall-gateway-api
+uninstall-gateway-api: go-mod-download-gateway-api
```
tao12345666333, line 286: why we need go-mod-download-gateway-api?

Makefile
```diff
@@ -258,3 +266,24 @@ kind-load-images:
             $(REGISTRY)/test-backend:dev \
             $(REGISTRY)/jmalloc/echo-server:dev \
             $(REGISTRY)/busybox:dev
+
+
+GATEWAY_API_PACKAGE ?= sigs.k8s.io/gateway-api
+GATEWAY_API_VERSION ?= v0.5.1
+GATEWAY_API_CRDS_LOCAL_PATH = $(shell go env GOPATH)/pkg/mod/$(GATEWAY_API_PACKAGE)@$(GATEWAY_API_VERSION)/config/crd
```
tao12345666333: If we already keep a copy of this CRD configuration in the our repo, why not just use this copy to install?

tao12345666333 (changes requested):

Makefile
AlinsRan: Dependent updates are convenient.

tao12345666333: but its uninstall-gateway-api

tao12345666333: Isn't it weird that I'm being asked to update it first when I want to delete it?

Makefile
AlinsRan: Dependent updates are convenient. The CRDs in the repo as an example, and maybe we can delete it?

Makefile
AlinsRan: This package does not exist when uninstalling is mainly avoided.

tao12345666333: This is obviously not necessary.

I can get the name of the CRD in the following two ways

```
kubectl api-resources --api-group=gateway.networking.k8s.io -o name
```

or

```
kubectl get crds  | awk '/gateway.networking.k8s.io/ {print $1}'
```

Then I can get the output:

```
gatewayclasses.gateway.networking.k8s.io
httproutes.gateway.networking.k8s.io
referencepolicies.gateway.networking.k8s.io
tcproutes.gateway.networking.k8s.io
tlsroutes.gateway.networking.k8s.io
udproutes.gateway.networking.k8s.io
```

Just need a `for` loop, you can delete these CRD resources directly.

This does not require any dependencies, even an offline environment.

AlinsRan: This is not very convenient. There is also a gateway-api webhook.

tao12345666333 (approved):

lingsamuel (approved):

Issue #592, opened 2021-07-14
proposal: Add IngressClass support for custom resources


tao12345666333:
Currently our custom resources are all namespace scoped.

If you want to use multiple apisix-ingress-controllers in the namespace at the same time, this is not allowed.

If we can add ingress class to custom resources, then we can let apisix-ingress-controller handle them separately.

Although this is a very low frequency scene. It is currently being discussed #578 

I think we may need to reach a consensus and add more complete information

tokers:
@tao12345666333 That would be a good way to support the soft isolation, but shall we still use the name `ingressClass`?

tao12345666333:
IngressClass is a general concept, I think it is more acceptable to take it.

Donghui0:
First: Ingress Class has a default value ("apisix"). If you continue to use ingressClass, then only ApisixRoute containing "apisix" will be listened by default. And if you want to listen to all ApisixRoute CRD by default, it seems impossible to achieve? Or put IngressClass To an empty string?
Second: Does ApisixRoute CRD need to add the Spec.IngressClass field?

tokers:
> First: Ingress Class has a default value ("apisix"). If you continue to use ingressClass, then only ApisixRoute containing "apisix" will be listened by default. And if you want to listen to all ApisixRoute CRD by default, it seems impossible to achieve? Or put IngressClass To an empty string?
> Second: Does ApisixRoute CRD need to add the Spec.IngressClass field?

A tricky way to support listen all ApisixRoute CRs is reserving a wildcard `*`, when the configured `ingressClass` is `*`, we listen all ApisixRoutes.

I think add `spec.IngressClass` is a good way.

gxthrj:
>First: Ingress Class has a default value ("apisix"). If you continue to use ingressClass, then only ApisixRoute containing "apisix" will be listened by default. And if you want to listen to all ApisixRoute CRD by default, it seems impossible to achieve? Or put IngressClass To an empty string?

The `IngressClass` can use `ingressclass.kubernetes.io/is-default-class` to set the default IngressClass.

>Second: Does ApisixRoute CRD need to add the Spec.IngressClass field? 

Yes, we need to add a field `spec.IngressClass`.

tao12345666333:
I agree add `spec.IngressClass` field

Donghui0:
[...]

PR #1523, opened 2022-12-15
feat: multiple controllers handle different ApisixRoute CRDs

Gallardot:
Signed-off-by: Gallardot <tttick@163.com>

Close https://github.com/apache/apisix-ingress-controller/issues/592

### Type of change:

- [ ] Bugfix
- [x] New feature provided
- [ ] Improve performance
- [ ] Backport patches

### Pre-submission checklist:


Gallardot:
I think this PR is ready for review. But there are some field naming issues that I think need to be discussed together.

1.  Is the new `instance-id` configuration appropriate?  The `ingress_class` configuration is not reused mainly to avoid version incompatibility. Reduce the cost of enabling this feature from inventory projects and reduce the hassle of version upgrades.

2. New `IngressClass` field in ApisixRoute object. Is this name appropriate? How about `gatewayclass`?

@tao12345666333 @AlinsRan @lingsamuel  @tokers  WDYT?

tao12345666333:
Thanks for your contributions. 
I will back tomorrow after some thinking 🤔

AlinsRan:
I think it depends on CRDs roadmap. But its evolution is not clear at present.
Maybe we can use annotations to temporarily transition. similar to ingress resource.

AlinsRan:
@tao12345666333  @lingsamuel @Gallardot  WDYT?

tao12345666333:
Sorry for the delay.  

I think using `IngressClass` is enough. [1]
In the resources of the Gateway API specification, there is `GatewayClass`, which allows for the definition of a controller name[2] to distinguish between different controllers.


So

- Using `controller-name`  instead of `instance-id`.   It already exists, make it configurable. https://github.com/apache/apisix-ingress-controller/blob/4241b673b3fd385e82d5ca3a39fed8acafa5bd64/pkg/config/config.go#L59-L61

- The `ingressClass` field is a good fit. We can reuse the IngressClass resource directly, and it is easier to integrate with existing tools or users' existing configuration.  No other resources need to be introduced.



1: https://kubernetes.io/docs/concepts/services-networking/ingress/#ingress-class
2: https://gateway-api.sigs.k8s.io/api-types/gatewayclass/#gatewayclass-controller-selection

lingsamuel:
ID is more like an auto generated string, but here it should be defined by the user. I agree with @tao12345666333

PR #1596, opened 2023-01-11
docs: add note about enabling the Plugin

### LINKED ISSUE #1564: [Docs] Add how to enable Plugins in the ApisixRoute#Plugins doc
The current guide mentions how you can configure Plugins in ApisixRoute resource but does not mention how to enable the Plugins that are not enabled in APISIX by default. A note on updating the values.yaml file during Helm install of APISIX could be mentioned as this is causing confusion: https://the-asf.slack.com/archives/CUC5MN17A/p1672147171239759

navendu-pottekkat:
Signed-off-by: Navendu Pottekkat <navendu@apache.org>

Fixes #1564 

Add a note mentioning how to enable the Plugins that are not enabled by default.

### DIFF
--- a/docs/en/latest/concepts/apisix_route.md
+++ b/docs/en/latest/concepts/apisix_route.md
@@ -178,6 +178,12 @@ spec:
 
 APISIX's [80+ Plugins](https://apisix.apache.org/docs/apisix/plugins/batch-requests/) can be used with APISIX Ingress. These Plugins have the same name as in the APISIX documentation.
 
+:::note
+
+If the Plugin is not enabled in APISIX by default, you can enable it by adding it to the `plugins` attribute in your `values.yaml` file while installing APISIX and Ingress controller via Helm.
+
+:::
+
 The example below configures [cors](https://apisix.apache.org/docs/apisix/plugins/cors/) Plugin for the route:
 
 ```yaml
--- a/docs/en/latest/concepts/apisix_route.md
+++ b/docs/en/latest/concepts/apisix_route.md
@@ -180,7 +180,7 @@ APISIX's [80+ Plugins](https://apisix.apache.org/docs/apisix/plugins/batch-reque
 
 :::note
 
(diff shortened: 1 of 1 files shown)


docs/en/latest/concepts/apisix_route.md
```diff
@@ -178,6 +178,12 @@ spec:
 
 APISIX's [80+ Plugins](https://apisix.apache.org/docs/apisix/plugins/batch-requests/) can be used with APISIX Ingress. These Plugins have the same name as in the APISIX documentation.
 
+:::note
+
+If the Plugin is not enabled in APISIX by default, you can enable it by adding it to the `plugins` attribute in your `values.yaml` file while installing APISIX and Ingress controller via Helm.
```
tao12345666333, line 183: I think we can also remind users to directly modify APISIX configuration files, what do you think?

navendu-pottekkat: Sure. We can add the line "You can also modify your APISIX configuration file directly (`conf/config.yaml`)."

tao12345666333: great!

navendu-pottekkat: Updated.

tao12345666333: Or if it is deployed in Kubernetes, it can be modified through configmap

tao12345666333 (approved):

docs/en/latest/concepts/apisix_route.md
navendu-pottekkat: Adding through ConfigMap would mean we are indirectly updating the `config.yaml` file right? Would it be confusing to the user if we mention it in a sentence? There are tutorials that mentions how a user can do that. We can mention these two methods as the easiest and the recommended ways to do that. A user can choose to go the ConfigMap path if they want. What do you think?

tao12345666333: ok, that make sense

juzhiyuan (approved):

PR #1628, opened 2023-01-31
docs: update the-hard-way doc contoller image version to 1.6.0

stubbornTanzhe:
### Type of change:

- [x] Update docs

if the image verion keeps to 1.4.0, the controller pod will CRASH, update it to 1.6.0

### DIFF
--- a/docs/en/latest/tutorials/the-hard-way.md
+++ b/docs/en/latest/tutorials/the-hard-way.md
@@ -322,7 +322,7 @@ spec:
     spec:
       containers:
         - name: apisix
-          image: "apache/apisix:2.15-alpine"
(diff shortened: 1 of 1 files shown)


tao12345666333:
Thank you for your contribution.
Did you follow the documentation exactly for installation with v1.6?

stubbornTanzhe:
> Thank you for your contribution. Did you follow the documentation exactly for installation with v1.6?

yes, i use 1.6.0 and it works. by the way, the apisix version also need to be updated to 2.15.0-alpine, which version cannot be found in dockerhub(I just forgot to add this for the first time)

stubbornTanzhe:
seems that the change ruined by my vim, i will give another PR

PR #1741, opened 2023-03-22
e2e: require certificate from file and add certificate generation script

### LINKED ISSUE #1182: e2e: require certificate from file and add certificate generation script
We should have a e2e certificates renew mechanics.

ps-19:
### Type of change:

- [ ] Bugfix
- [ ] New feature provided
- [x] Improve performance
- [ ] Backport patches
- [ ] Documentation
- [ ] Refactor
- [ ] Chore
- [x] CI/CD or Tests

This PR adds functionality to require a certificate from a file, and also includes a certificate generation script. The purpose of this is to simplify the process of generating and using certificates in the application, by providing an automated way to generate new certificates as needed.

With this PR, users of the application can simply provide the path to an existing certificate file, rather than having to generate a new certificate manually each time. In addition, the included certificate generation script makes it easy for developers to generate new certificates as needed, without having to manually create and configure them.

Overall, this PR makes it easier to work with SSL/TLS certificates in the application, by simplifying the process of generating and using certificates. It provides an automated and standardized way to handle certificates, making the application more secure and reliable.

Fixes #1182

### Pre-submission checklist:


tao12345666333:
@ps-19 please check the CI erros.

lingsamuel:
The issue was intended to solve two problems: the large number of raw certificates in e2e is hard to read and understand, and there is no renewal mechanism.
What I want is to add the renewal mechanism, which is equivalent to being able to dynamically generate new certificates every time you run e2e, so that the reader can understand the configuration by read the certificates' generation parameters.

Therefore, instead of simply splitting the certificates into files, you should modify the CI workflow to dynamically generate these certificates.
Also, the `keyCompare` should be modified too.

ps-19:
Can I use cert-manager?

```

apiVersion: cert-manager.io/v1alpha2
kind: Certificate
metadata:
  name: example-cert
spec:
  secretName: example-tls
  duration: 2160h
  renewBefore: 360h
  commonName: api7.com
  dnsNames:
    - api7.com
  issuerRef:
    name: letsencrypt
    kind: ClusterIssuer

```

ps-19:
https://cert-manager.io/docs/reference/api-docs/

ps-19:
kubectl apply -f certificate.yaml

kubectl get secret example-tls -o jsonpath='{.data.tls\.crt}' | base64 -d

lingsamuel:
We don't need such a heavy component just for generating some simple test certificates. `openssl` can do everything.

Issue #1760, opened 2023-03-31
bug: apisix restarts, ingress controller unavailable


Donghui0:
### Current Behavior

The following scenario:
1 * apisix gateway instance. 
2 * ingress controller instances.

Apisix gateway restart occurred,   the health check interface of leader ingress controller  throws 500 error code.
interface: /apisix/healthz
500 error msg: {"status":"timed out waiting for the condition"}
This interface keeps reporting errors, but another candidate ingress controller interface is normal.

### Expected Behavior

It should be able to recover automatically after reporting an error for a period of time.
At present, it can only be solved by restarting the ingress controller

### Error Logs

 curl http://10.227.55.111:8080/apisix/healthz
{"status":"timed out waiting for the condition"}

### Steps to Reproduce

resource preparation:
1 * apisix gateway instance. 
2 * ingress controller instances.

step1: restart apisix gateway.
step2: Observe the leader ingress controller status and logs.

### Environment

- APISIX Ingress controller version: v1.5.0
- Kubernetes cluster version (run `kubectl version`): v1.22

basefas:
I want to work on this issue please assign this to me.

tao12345666333:
Assigned, thanks

PR #1779, opened 2023-04-11
fix: health check bug

basefas:
Keep health checker running when health check failed. 
Make health check function pure.

### Type of change:

Bugfix

for #1760

### DIFF
--- a/pkg/providers/controller.go
+++ b/pkg/providers/controller.go
@@ -570,16 +570,20 @@ func (c *Controller) checkClusterHealth(ctx context.Context, cancelFunc context.
 
 		err := c.apisix.Cluster(c.cfg.APISIX.DefaultClusterName).HealthCheck(ctx)
 		if err != nil {
-			// Finally failed health check, then give up leader.
-			log.Warnf("failed to check health for default cluster: %s, give up leader", err)
 			c.apiServer.HealthState.Lock()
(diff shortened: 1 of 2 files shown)


pkg/providers/controller.go
```diff
@@ -570,16 +570,20 @@ func (c *Controller) checkClusterHealth(ctx context.Context, cancelFunc context.
...
+			// Finally failed health check, then give up leader.
+			log.Warnf("failed to check health for default cluster: %s, give up leader", err)
+		} else {
+			if c.apiServer.HealthState.Err != nil {
+				c.apiServer.HealthState.Lock()
+				c.apiServer.HealthState.Err = err
+				c.apiServer.HealthState.Unlock()
+			}
+			log.Debugf("success check health for default cluster")
+			c.MetricsCollector.IncrCheckClusterHealth(c.name)
```
tao12345666333, line 585: The code here can be simplified. The state change for HealthState does not need to be placed in an if/else statement.

basefas: the code in if block because it will only be changed when current health state error is nil and last health state error is not nil. Otherwise，it will be changed every health check time.

tao12345666333 (approved):

lingsamuel (approved):

AlinsRan (approved):

PR #1780, opened 2023-04-11
fix: secret reference update Ingress

### LINKED ISSUE #1777: bug: Deleting secret when using ingress tls will cause controller to crash
### Current Behavior

When using the tls of the ingress resource, deleting the secret will cause the controller to crash.

2edd6f1b0b7873f739627d10085d06b

The code that caused the crash:

https://github.com/apache/apisix-ingress-controller/blob/f9aba3945bef2a385ff8261d42e30599b144023c/pkg/providers/ingress/ingress.go#L579-L586

Ingress example:
```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: httpbin-ingress-https
spec:
  ingressClassName: apisix
  tls:
  - hosts:
    - example.com
    secretName: example-secret
  rules:
  - host: example.com
    http:
      paths:
      - path: /
        pathType: Prefix
        backend:
          service:
            name: httpbin
            port:
              number: 80
```

### Expected Behavior

It work.

### Error Logs

_No response_

### Steps to Reproduce

1. Deploy apisix-ingress-controller
2. Create a secret `example-secret`
3. Create a ingress tls with secret
```yaml
spec:
  tls:
  - hosts:
    - example.com
    secretName: example-secret
```
4. Delete secret `example-secret`

### Environment

- APISIX Ingress controller version (run `apisix-ingress-controller version --long`) 1.6.1
- Kubernetes cluster version (run `kubectl version`) 1.26.3
- OS version if running APISIX Ingress controller in a bare-metal environment (run `uname -a`) Debian

lingsamuel:
### Type of change:

- [x] Bugfix

Fixes #1777

Fix a bug that causes nil reference and secret change won't trigger Ingress resource sync.

### Pre-submission checklist:


pkg/providers/ingress/ingress.go
```diff
@@ -131,7 +130,7 @@ func (c *ingressController) sync(ctx context.Context, ev *types.Event) error {
 	}
 
 	if err != nil {
-		if !k8serrors.IsNotFound(err) {
+		if !k8serrors.IsNotFound(err) || (k8serrors.IsNotFound(err) && ev.Type != types.EventDelete) {
```
AlinsRan, line 133: There are two questions here
1. Events are not always up-to-date. We always retrieve the latest object, which may have been deleted. This logic can lead to infinite retries.
2. Redundant logic, same as line 138.

image

```go
if !k8serrors.IsNotFound(err) || (k8serrors.IsNotFound(err) && ev.Type != types.EventDelete) { return }
if ev.Type != types.EventDelete { return } # Redundant 

<==>

if !k8serrors.IsNotFound(err) { return }
if ev.Type != types.EventDelete { return }

```

pkg/providers/ingress/ingress.go
```diff
@@ -152,6 +152,36 @@ func (c *ingressController) sync(ctx context.Context, ev *types.Event) error {
 		}
 		ing = ev.Tombstone.(kube.Ingress)
 	}
+	if ing == nil {
```
AlinsRan, line 155: Same reason as above.
```golang
return nil
```

AlinsRan (changes requested):

AlinsRan:
When deleting a secret, does the SSL object of APISIX exist?
Missing e2e case for delete-secret?

tao12345666333 (approved):

AlinsRan (approved):

PR #1806, opened 2023-04-21
Controller restarted

ps-19:
### Type of change:

- [ ] Bugfix
- [x] New feature provided
- [ ] Improve performance
- [ ] Backport patches
- [ ] Documentation
- [ ] Refactor
- [ ] Chore
- [ ] CI/CD or Tests

The pull request proposes to add a feature to the end-to-end (e2e) test cases of a system to check if the controller has restarted. The goal is to detect and report any issues with the controller that may cause it to crash and restart, which could lead to false-positive results in the test cases.

The proposed solution is to add a detection mechanism at the end of each e2e use case that checks if the controller has restarted. This check will be performed in the afterEach() function and will report any restarts detected during the test case cleanup. The code snippet provided in the pull request suggests using the Kubernetes API to retrieve the status of the controller deployment and comparing the number of replicas specified in the deployment specification with the number of ready replicas.

If the check detects a controller restart, it will fail the test case and return true, indicating that a restart was detected. Otherwise, it will return false, indicating that no restart was detected.

### Pre-submission checklist:

### DIFF
--- a/test/e2e/scaffold/scaffold.go
+++ b/test/e2e/scaffold/scaffold.go
@@ -30,6 +30,7 @@ import (
 	"strings"
 	"text/template"
 	"time"
+	"encoding/json"
 
 	"github.com/apache/apisix-ingress-controller/pkg/config"
 	"github.com/apache/apisix-ingress-controller/pkg/log"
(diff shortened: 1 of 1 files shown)


ps-19:
Consider This A Professional Courtesy.

PR #1856, opened 2023-06-11
docs: Update issuer.yaml

harvies:
error: resource mapping not found for name: "ca-issuer" namespace: "" from "issuer.yaml": no matches for kind "Issuer" in version "cert-manager.io/v1alpha2"
ensure CRDs are installed first

https://cert-manager.io/docs/installation/upgrading/remove-deprecated-apis/


tao12345666333 (approved):
Thanks

lingsamuel (approved):

PR #1983, opened 2023-09-26
fix: ginko versions in e2e tests

daroga0002:
### Type of change:

- [ ] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches
- [ ] Documentation
- [ ] Refactor
- [ ] Chore
- [x] CI/CD or Tests

This is solving issue with hardcoded version of Ginkgo module. There is strict dependency between tests in go using ginkgo package and local Ginkgo binary used for those tests. Currently after each update in dependencies CI start failing with log:

```sh
Ginkgo detected a version mismatch between the Ginkgo CLI and the version of Ginkgo imported by your packages:
  Ginkgo CLI Version:
    2.12.0
  Mismatched package versions found:
    2.12.1 used by e2e

  Ginkgo will continue to attempt to run but you may see errors (including flag
  parsing errors) and should either update your go.mod or your version of the
  Ginkgo CLI to match.

  To install the matching version of the CLI run
    go install github.com/onsi/ginkgo/v2/ginkgo
  from a path that contains a go.mod file.  Alternatively you can use
    go run github.com/onsi/ginkgo/v2/ginkgo
  from a path that contains a go.mod file to invoke the matching version of the
  Ginkgo CLI.

  If you are attempting to test multiple packages that each have a different
  version of the Ginkgo library with a single Ginkgo CLI that is currently
  unsupported.
```

This PR should fetch a correct version from e2e tests go.mod file and then install it in correct version.

### Pre-submission checklist:

### DIFF
--- a/.github/workflows/e2e-test-ci-v2-cron-dev.yml
+++ b/.github/workflows/e2e-test-ci-v2-cron-dev.yml
@@ -165,7 +165,8 @@ jobs:
 
       - name: Install ginkgo
         run: |
-          go install github.com/onsi/ginkgo/v2/ginkgo@v2.12.0
+          GINKGO_VERSION=$(cd test/e2e && go list -m  -mod=readonly  -f {{.Version}}  github.com/onsi/ginkgo/v2)
+          go install github.com/onsi/ginkgo/v2/ginkgo@$GINKGO_VERSION
           sudo cp ~/go/bin/ginkgo /usr/local/bin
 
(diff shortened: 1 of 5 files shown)


daroga0002:
It should be working correctly (locally tested and it is ok), but we will see when workflows will run what are results

Gallardot (approved):

PR #2012, opened 2023-10-18
fix: plugin_metadata not populating to the correct etcd key

zhuoyang:
### Type of change:
- [x] Bugfix

Fix `plugin_metadata_cm` not working due to populating to wrong etcd key
Before this fix, this can be seen in the debug log
```
2023-10-18T11:31:03+08:00	[35mdebug[0m	apisix/cluster.go:1225	push event to adapter	{"event": "create", "key": "/apisix/plugin_metadatas/datadog", "value": "{\"host\":\"datadog.datadog\",\"namespace\":\"apisix\",\"port\":8125}"}
```

### Pre-submission checklist:

### DIFF
--- a/pkg/apisix/plugin_metadata.go
+++ b/pkg/apisix/plugin_metadata.go
@@ -178,7 +178,7 @@ type pluginMetadataMem struct {
 
 func newPluginMetadataMem(c *cluster) PluginMetadata {
 	return &pluginMetadataMem{
-		url:      c.baseURL + "/plugin_metadatas",
+		url:      c.baseURL + "/plugin_metadata",
 		resource: "plugin_metadata",
 		cluster:  c,
 	}
--- a/pkg/apisix/plugin_metadata.go
+++ b/pkg/apisix/plugin_metadata.go
@@ -179,7 +179,7 @@ type pluginMetadataMem struct {
 func newPluginMetadataMem(c *cluster) PluginMetadata {
 	return &pluginMetadataMem{
 		url:      c.baseURL + "/plugin_metadatas",
-		resource: "plugin_metadatas",
+		resource: "plugin_metadata",
 		cluster:  c,
 	}
 }


lingsamuel (approved):

zhuoyang:
I agree, but I will need some time as there are other priority for me right now

pkg/apisix/plugin_metadata.go
```diff
@@ -179,7 +179,7 @@ type pluginMetadataMem struct {
 func newPluginMetadataMem(c *cluster) PluginMetadata {
 	return &pluginMetadataMem{
 		url:      c.baseURL + "/plugin_metadatas",
```
AlinsRan, line 181:
```suggestion
		url:      c.baseURL + "/plugin_metadata",
```

indrekj:
I saw you fixed the wrong path in [the helm charts](https://github.com/apache/apisix-helm-chart/pull/646). Do you think it's also worth fixing it in this repo as well: https://github.com/apache/apisix-ingress-controller/blob/master/conf/config-default.yaml#L86 (should also be on the root level)

pkg/apisix/plugin_metadata.go
zhuoyang: I saw this [code](https://github.com/apache/apisix/blob/3b16fce060fbac39ee054784213a6bc81df526fc/apisix/control/v1.lua#L468) in the apisix repo, is it related?

AlinsRan: pls refs: https://apisix.apache.org/docs/apisix/admin-api/#plugin-metadata

Actually, the `plugin_metadata` is the correct one

AlinsRan: The code you mentioned is the control API
https://apisix.apache.org/docs/apisix/control-api/#get-v1plugin_metadatas.

zhuoyang: I see, I actually don't quite understand what is the url used for, why in my testing even though the url is supposedly wrong but it's still working ? As you are the original committer of the code could you briefly explain to me for my learning purpose?

zhuoyang:
I am actually having a hard time understanding the e2e test, it seems like there's [e2e test](https://github.com/apache/apisix-ingress-controller/blob/master/test/e2e/suite-ingress/suite-ingress-resource/configmap.go) written for this feature already but why does it not fail in the first place? 
I did pass in `ENABLED_ETCD_SERVER=true` when I run `make e2e-test-local`, tried both with and without my fix, the test still pass anyway

pkg/apisix/plugin_metadata.go
AlinsRan: It only represents simulating the corresponding `plugin_metadata` etcd key here, but in fact, it does not have many uses.

AlinsRan (approved):

lingsamuel (approved):

zhuoyang:
thanks for merging! is it possible to release a hotfix for this PR?

lingsamuel:
> thanks for merging! is it possible to release a hotfix for this PR?

Sure. Patch version 1.7.1 is voting, thread here:

https://lists.apache.org/thread/fjy8t21nrj2yyfvrmyshlokc6w3jcgp

PR #2065, opened 2023-11-22
fix: make regex for hosts consistent with APISIX

### LINKED ISSUE #2064: bug:  Host matching cannot carry ports
### Current Behavior

The ApisixRoute "rbd-http-routes" is invalid: spec.http.match.hosts: Invalid value: "8.130.141.31:7070": spec.http.match.hosts in body should match '^\*?[0-9a-zA-Z-._]+$'

```js
/^\*?[0-9a-zA-Z-._]+$/.test('8.130.141.31:7070')
false 
```

```js
/^\*?[0-9a-zA-Z-._]+$/.test('8.130.141.31') 
true
```

Some users access their applications through IP and ports, and use the HTTP protocol, so they hope to increase support for the HTTP port

Revolyssup:
### Type of change:
Fixes #2064 
Currently there is a mismatch between the regex that APISIX allowed[1] and what ApisixRoute CRD allowed. This PR makes sure that both are consistent.
More specifically - Both `1.2.3.4` and `1.2.3.4:443` should be considered valid `hosts` field

1. https://github.com/apache/apisix-ingress-controller/blob/ea6fbc95d774d9517aded455b59ce238c06c1edf/conf/apisix-schema.json#L961 

- [ ] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches
- [ ] Documentation
- [ ] Refactor
- [ ] Chore
- [ ] CI/CD or Tests

### Pre-submission checklist:

### DIFF
--- a/samples/deploy/crd/v1/ApisixRoute.yaml
+++ b/samples/deploy/crd/v1/ApisixRoute.yaml
@@ -110,7 +110,7 @@ spec:
                             minItems: 1
                             items:
                               type: string
-                              pattern: "^\\*?[0-9a-zA-Z-._]+$"
+                              pattern: "^\*?[0-9a-zA-Z-._\[\]:]+$"
                           methods:
                             type: array
                             minItems: 1


AlinsRan:
Is there a problem with the admin-api mode?
We need test cases to cover it.

Revolyssup:
> Is there a problem with the admin-api mode? We need test cases to cover it.

The admin api is fine. The regex of the same field that admin api expects and what ingress controller expects is different. So this PR makes sure that regex is same for the same field.
Okay I'll add a test case.

Revolyssup:
This issue needs to be fixed first for the CI to pass here https://github.com/api7/lua-resty-radixtree/issues/138

test/e2e/suite-features/route_match_exprs.go
```diff
@@ -79,6 +79,57 @@ spec:
...
+apiVersion: apisix.apache.org/v2
+kind: ApisixRoute
+metadata:
+ name: httpbin-route
+spec:
+ http:
+ - name: rule1
+   match:
+     hosts:
+     - httpbin.org
```
AlinsRan, line 95: Does this use case seem unrelated to PR? What about the port?

shreemaan-abhishek (approved):

AlinsRan (approved):

PR #2139, opened 2024-01-17
refactor(cmd/ingress): invert signal ctx logic

acuteaura:
### Type of change:

- [x] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches
- [ ] Documentation
- [x] Refactor
- [ ] Chore
- [ ] CI/CD or Tests

(verbatin commit message)

this commit changes the signal handling in cmd/ingress to be wrapped in a context, and inverts which goroutine runs the controller and which watches for the context to be cancelled, which allows some scaffolding (`sync.WaitGroup`) to be removed and now properly handles the controller exiting with `nil` (as it does when leader election fails)

### DIFF
--- a/cmd/ingress/ingress.go
+++ b/cmd/ingress/ingress.go
@@ -50,6 +50,7 @@ func contextWithSignalCancel(ctx context.Context, signals ...os.Signal) context.
 
 		sig := <-sigCh
 		log.Infof("signal %d (%s) received", sig, sig.String())
+		signal.Stop(sigCh)
 		close(sigCh)
 		cancel()
 	}()
--- a/cmd/ingress/ingress.go
+++ b/cmd/ingress/ingress.go
@@ -43,7 +43,7 @@ func dief(template string, args ...interface{}) {
 }
 
 func contextWithSignalCancel(ctx context.Context, signals ...os.Signal) context.Context {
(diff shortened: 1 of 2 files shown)


Revolyssup:
@acuteaura Can you fix the failing unit test? Rest, the change LGTM

acuteaura:
The test failing is code I didn't touch, which is just an unstable test, it seems to be iterating over a map with two elements and not accounting for map iteration order not being stable.

> The iteration order over maps is not specified and is not guaranteed to be the same from one iteration to the next.

https://go.dev/ref/spec#RangeClause

The range over map is here:

https://github.com/acuteaura/apisix-ingress-controller/blob/refactor/root-context-signaling/pkg/types/labels.go#L45

Should go into another PR, and I'd be happy to rebase on that.

Revolyssup:
@acuteaura Actually other than the flaky test, this one is also failing
Can you take a look? https://github.com/apache/apisix-ingress-controller/actions/runs/7555667043/job/20609448277

acuteaura:
Ah, right, Notify can fire more than once, should be fixed now.

Revolyssup (approved):

AlinsRan (approved):

Revolyssup:
@acuteaura Thanks for your contribution. You can now create the other PR as well

PR #1626, opened 2023-01-30
Epnotfound

goblain:
### Type of change: 

Bugfix

### What this PR does / why we need it:

Solves #1625

Issue #810, opened 2021-12-23
bug: service not found，and  then I edited it and it still kept reporting errors.


GhangZh:
### Issue description

My original configuration of ApisixRoute had a service that didn't exist, and then I edited ApisixRoute and it still kept reporting errors.
![image](https://user-images.githubusercontent.com/92301646/147242908-1f215777-70ac-4710-8c01-9655e6d792af.png)
![image](https://user-images.githubusercontent.com/92301646/147243156-af89b78f-6847-4866-a7df-f8608291f39c.png)

### Environment

- your apisix-ingress-controller version (output of apisix-ingress-controller version --long):1.3.0
- your Kubernetes cluster version (output of kubectl version):1.18
- if you run apisix-ingress-controller in Bare-metal environment, also show your OS version (uname -a):Linux

### Minimal test code / Steps to reproduce

kubectl edit ar
kubectl logs -f apisix-ingress-controller 

### Actual result

Successful

### Error log

failed to translate old ApisixRoute service not found

tao12345666333:
Thanks for report. I will try to reproduce it

gxthrj:
Do you mean that after deleting the `service` that does not exist in `ApisixRoute`, `ingress-controller` is still converting the old `service`?

gxthrj:
This is indeed a bug.
When the `ApisixRoute` resource is updated, the old `ApisixRoute` resource object is indeed recorded, so this wrong object always exists.

GhangZh:
> This is indeed a bug. When the `ApisixRoute` resource is updated, the old `ApisixRoute` resource object is indeed recorded, so this wrong object always exists.

At the moment I can only restart the ingress-controller after updating the service pod?

PR #2146, opened 2024-01-22
fix: retry syncing when the error is k8s.NotFound

Revolyssup:
### Type of change:

The flow is that first `sync` function is called which returns and error. Then that error is sent to `handleSyncErr` function.

For all the resources, the case of error where the resource object is not found is handled separately even before handleSyncErr function is called. So usually when NotFound error is caught in `handleSyncErr` function, the NotFound error is not for that resource for which the controller is running but for some other resource. In that case, a Retry should be made. 

Example, For ApisixRoute resource, the NotFound is already handled [here](https://github.com/apache/apisix-ingress-controller/blob/0385d285040b77d7fa10dd3610f90b15b87ed4d2/pkg/providers/apisix/apisix_route.go#L292) where the logic is to ignore the error. 

Here is the bug:
Let's say we create Route before the Service. So the first time, we will get an error which is "Endpoint not found". But later endpoint could be created, so this event should not be discarded yet. The current logic in handleSyncErr will discard the event mistaking that to be "ApisixRoute not found" and even after the endpoint is created, the route will not be created.

How this bug was not found so far:
We currently wrap the k8sNotFound error in our own custom error, so when it bubbled up to handleSyncErr, it was not detected as a k8sNotFound error and the retry was triggered anyways.

How the bug was detected:
While fixing the aforementioned issue, the wrapping error was taken off to add the logic that when the error is k8sNotFound, we don't spam the user. While doing so, the wrapping was removed from a deeply nested function which allowed k8sNotFound to bubble up finally and then according to the current logic, it was discarded. Even when the NotFound was not for the Route resource but Endpoint resource.

Replica of https://github.com/apache/apisix-ingress-controller/pull/1626

- [ ] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches
- [ ] Documentation
- [ ] Refactor
- [ ] Chore
- [ ] CI/CD or Tests

### Pre-submission checklist:


Vacant2333 (approved):

pkg/providers/translation/service.go
```diff
@@ -35,9 +35,15 @@ import (
...
 	endpoint, err := t.EndpointLister.GetEndpoint(namespace, name)
 	if err != nil {
-		return nil, &TranslateError{
-			Field:  "endpoints",
-			Reason: err.Error(),
+		if k8serrors.IsNotFound(err) {
+			//Do not error wrap so that IsNotFound error can be handled separately on upper level
+			//Reference https://github.com/apache/apisix-ingress-controller/issues/1625
+			return nil, err
+		} else {
```
zll600, line 42: It seems that this `else` is not necessary. Can you remove it?

zll600: But if the error is `NotFound`, this function will return immediately. Otherwise this function will return a `&TranslateEror`. right?

pkg/providers/apisix/apisix_cluster_config.go
```diff
@@ -219,15 +219,6 @@ func (c *apisixClusterConfigController) handleSyncErr(obj interface{}, err error
 		c.MetricsCollector.IncrSyncOperation("clusterConfig", "success")
 		return
 	}
-	event := obj.(*types.Event)
```
AlinsRan, line 222: refs: https://github.com/apache/apisix-ingress-controller/issues/810

1. There may be an expired event during the retry, which means it was deleted when the retry occurred, and this event will continue to be triggered.
2. What is the reason why this resource status needs to be retried? What should be the retry strategy? I think rough deletion is not advisable.
3. The changes here are not related to the title of the PR. If you think you need to try again, please submit a separate PR for discussion

pkg/providers/apisix/apisix_consumer.go
```diff
@@ -184,15 +184,6 @@ func (c *apisixConsumerController) handleSyncErr(obj interface{}, err error) {
 		c.MetricsCollector.IncrSyncOperation("consumer", "success")
 		return
 	}
-	event := obj.(*types.Event)
```
AlinsRan, line 187: dito

AlinsRan (changes requested):

PR #2152, opened 2024-01-31
refactor: simplify leader election

acuteaura:
mostly based on the upstream example:
https://github.com/kubernetes/client-go/blob/master/examples/leader-election/main.go

this also incidentally fixes bugs where `run` returns nil (i.e. when failing to reach the Admin API) by causing the context to cancel via defer.

- [x] Bugfix
- [x] Refactor

### DIFF
--- a/pkg/providers/controller.go
+++ b/pkg/providers/controller.go
@@ -213,7 +213,7 @@ func (c *Controller) Run(ctx context.Context) error {
 					zap.String("pod", c.name),
 				)
 				// rootCancel might be to slow, and controllers may have bugs that cause them to not yield
-				// the safest way to step down is to simply cause a pod restart, making
+				// the safest way to step down is to simply cause a pod restart
 				os.Exit(0)
 			},
 		},
@@ -226,7 +226,7 @@ func (c *Controller) Run(ctx context.Context) error {
 	if leaderElectionConfig.WatchDog != nil {
 		leaderElectionConfig.WatchDog.SetLeaderElection(leaderElector)
 	}
-	// todo: this should never be neccessary
+	// todo: this should never be necessary if only one controller runs
 	c.elector = leaderElector
 
 	leaderElector.Run(rootCtx)
--- a/pkg/providers/controller.go
+++ b/pkg/providers/controller.go
@@ -18,7 +18,6 @@ import (
 	"context"
 	"fmt"
 	"os"
-	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


acuteaura:
some remarks on this PR:

* a bunch of code seems to treat leader status as "skipping writes" (via the `Elector`). this is pretty hard to reason about and prone to errors, so this PR changes it to a more traditional standby (with informers warmed up) that doesn't run any controllers.
* alongside that, the controller now hard exits when it loses leader status.
* `run` got an error return value, because a lot of error conditions were just silently unhandled. A future PR should improve these errors with some kind of wrapping and remove explicit logging in `run` itself so it all bubbles up
* I did not touch the API server, because I'm not 100% about all the things it does, but we should consider failing a readiness probe if it is not leader if code in there relies on controllers running - though it did not have access to `Elector` until this point either.

Revolyssup:
@acuteaura Can you fix the merge conflicts?

Revolyssup:
@acuteaura Unit tests failing

pkg/providers/controller.go
```diff
@@ -227,39 +208,30 @@ func (c *Controller) setupLeaderElection() error {
...
 			OnStoppedLeading: func() {
 				c.MetricsCollector.ResetLeader(false)
-				log.Infow("controller now is running as a candidate",
+				log.Infow("controller lost leader, exiting",
 					zap.String("namespace", c.namespace),
 					zap.String("pod", c.name),
 				)
+				// rootCancel might be to slow, and controllers may have bugs that cause them to not yield
+				// the safest way to step down is to simply cause a pod restart
+				os.Exit(0)
```
Revolyssup, line 217: Makes sense

Revolyssup:
an e2e test related to leader election was also failing. I reran it

acuteaura:
@Revolyssup could you have a look at this failed test? I'm not really sure how it could fail in the beforeEach step when that's all shared among tests. Or maybe re-run the tests to make sure this isn't a flake.

It's unfortunately not so trivial to run this test locally, but I'm getting to it...

zll600:
ping @Revolyssup

bstasz:
Hello,
@acuteaura Would you reopen this PR so that we can have new test result logs? I would like to look into this and possibly fix it.

acuteaura:
I can't, but I can rebase and re-submit later.

Issue #2167, opened 2024-02-19
bug:  plugin config "concurrent map writes" 


liushv0:
### Current Behavior

controller processor occasionally crash during running,
cause by "fatal error: concurrent map writes"

### Expected Behavior

no panic

### Error Logs

<img width="1443" alt="image" src="https://github.com/apache/apisix-ingress-controller/assets/19424192/1d8b851e-6f55-4fed-8496-8c689faf3782">

### Steps to Reproduce

function call stack like this:
<img width="2700" alt="apisix-ingress-controller start" src="https://github.com/apache/apisix-ingress-controller/assets/19424192/741d36c0-db0c-4a0f-ae7e-a4a84d7848d8">

so function   Plugins.DeepCopyInto(pkg/types/apisix/v1/types.go:187)  could be called by multiple goroutine at same time,
type  Plugins  is alias name  to map， 
it is not concurrent safe，this will cause  ”concurrent map writes“  error

### Environment

- APISIX Ingress controller version  1.7.1

Revolyssup:
The struct on which unmarshal is being performed is created with new() everytime when DeepCopy is called. Which means the data which is being written is not shared between function calls, so how can there be concurrent writes? I am trying to understand that right now.

lkad:
"I also encountered the same bug; the ingress can restart about three times a week."

PR #2168, opened 2024-02-19
fix: concurrent map write cause by concurrent invocation of function …

liushv0:
…DeepCopy (#2167)

### Type of change:

- [x] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches
- [ ] Documentation
- [ ] Refactor
- [ ] Chore
- [ ] CI/CD or Tests

### Pre-submission checklist:

### DIFF
--- a/pkg/apisix/cache/memdb.go
+++ b/pkg/apisix/cache/memdb.go
@@ -17,6 +17,7 @@ package cache
 
 import (
 	"errors"
+	"sync"
 
 	"github.com/hashicorp/go-memdb"
 
@@ -31,7 +32,16 @@ var (
 )
 
 type dbCache struct {
-	db *memdb.MemDB
+	db                      *memdb.MemDB
+	routeLock               sync.Mutex
+	sslLock                 sync.Mutex
+	upstreamLock            sync.Mutex
+	streamRouteLock         sync.Mutex
+	globalRuleLock          sync.Mutex
+	consumerLock            sync.Mutex
+	schemeLock              sync.Mutex
+	pluginCfgLock           sync.Mutex
+	upstreamSrvRelationLock sync.Mutex
 }
 
 // NewMemDBCache creates a Cache object backs with a memory DB.
@@ -97,6 +107,8 @@ func (c *dbCache) GetRoute(id string) (*v1.Route, error) {
 	if err != nil {
 		return nil, err
 	}
+	c.routeLock.Lock()
+	defer c.routeLock.Unlock()
 	return obj.(*v1.Route).DeepCopy(), nil
 }
 
@@ -105,6 +117,8 @@ func (c *dbCache) GetSSL(id string) (*v1.Ssl, error) {
 	if err != nil {
 		return nil, err
 	}
+	c.sslLock.Lock()
+	defer c.sslLock.Unlock()
 	return obj.(*v1.Ssl).DeepCopy(), nil
 }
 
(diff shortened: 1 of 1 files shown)


Revolyssup:
The struct on which unmarshal is being performed is created with new() everytime when DeepCopy is called. Which means the data which is being written is not shared between function calls, so how can there be concurrent writes? I am trying to understand that right now.

Revolyssup:
Because we have the same kind of implementation as there are in kubernetes repos and nowhere do they use a lock 
https://github.com/search?q=org%3Akubernetes%20DeepCopy(&type=code

Revolyssup:
@liushv0 Can you answer the above questions to help me review this PR?

liushv0:
image


sorry for late  :( 

apisix-route sync goroutine（gr1）will get test-ar from MemDB，
apisix-plugin-config sync goroutine （gr2 ）will get test-apc from MemDB，
test-apc has be refered by test-ar。


now  gr1 begin to sync CR test-apc，func `PluginConfig.DeepCopyInto` will be called，in this function，`Plugins#DeepCopyInto` be called。
image



at the same time， gr2 begin to sync CR test-apc，func  `Route#DeepCopyInto`  will be called，in this function，`Plugins#DeepCopyInto`  also be called。 
image


Plugins in test-apc and Plugins in test-ar  pointer to same instance，so two goroutine concurrent  marshal  this Plugins instance，and panic

Revolyssup (approved):

Revolyssup:
@liushv0 I think there is no point in having locks as well as DeepCopy. I think this line need to be removed https://github.com/apache/apisix-ingress-controller/blob/58aab33bc6d215c6a3f826daedbd5a5d85ca6c23/pkg/types/apisix/v1/zz_generated.deepcopy.go#L336. This line may have been added when this file was autto generated.

Revolyssup:
@liushv0 Are you available to make changes here?

pkg/apisix/cache/memdb.go
```diff
@@ -113,6 +127,8 @@ func (c *dbCache) GetUpstream(id string) (*v1.Upstream, error) {
 	if err != nil {
 		return nil, err
 	}
+	c.upstreamLock.Lock()
+	defer c.upstreamLock.Unlock()
```
Revolyssup, line 131: @liushv0 I think there is no point in having locks as well as DeepCopy. I think this line need to be removed https://github.com/apache/apisix-ingress-controller/blob/58aab33bc6d215c6a3f826daedbd5a5d85ca6c23/pkg/types/apisix/v1/zz_generated.deepcopy.go#L336. This line may have been added when this file was autto generated.

Revolyssup (changes requested):

pkg/apisix/cache/memdb.go
liushv0: I guess remove  this line is effective，and less change，
but honesty，I am not exactly understand  the intention of this line，set *out = *in for what。
other side，this line  is generated by code-gen tools，maybe somebody will rerun  code-gen tools，and overwrite the  modification。
that is why I choose use  locks。

Revolyssup:
@liushv0 Can you modify this PR to remove locks and instead apply this suggestion?https://github.com/apache/apisix-ingress-controller/pull/2168#discussion_r1527877936

lkad:
i have meet same bug ,has anyone can fix  it ?

lkad:
@liushv0  你这段代码在自己环境中使用了吗，我目前也碰到同样的问题, 烦请有空告知下，非常感谢.

PR #2208, opened 2024-04-11
fix: remove direct pointer dereference to avoid concurrent usage

Revolyssup:
### Type of change:
Context: https://github.com/apache/apisix-ingress-controller/pull/2168#issuecomment-1996612561

- [x] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches
- [ ] Documentation
- [ ] Refactor
- [ ] Chore
- [ ] CI/CD or Tests

### Pre-submission checklist:


AlinsRan:
When generating code, it will overwrite the original content, which is not a good solution.

Revolyssup:
> When generating code, it will overwrite the original content, which is not a good solution.

Yeah. How do you think we can get around this? 
```go
*out = *in
```
causes these structs internal pointers pointing to same addresses before the code afterwards can run and perform deepcopy. This is causing the concurrent access.

Issue #1980, opened 2023-09-26
bug: ingress-controller doesn't recover from failed sync


acuteaura:
### Current Behavior

occasionally, our controller will stop doing anything after failing to sync, but it also doesn't exit

we have logs for this occurring immediately following an upgrade, but we've seen similar behavior elsewhere too

### Expected Behavior

The ingress controller should either retry running its logic or exit.

### Error Logs

```
2023-09-22T17:40:34+08:00	info	ingress/ingress.go:145	start ingress controller
2023-09-22T17:40:34+08:00	info	providers/controller.go:155	start api server
2023-09-22T17:40:34+08:00	info	providers/controller.go:170	start leader election
I0922 17:40:34.168741       1 leaderelection.go:250] attempting to acquire leader lease apisix/ingress-apisix-internal-leader...
2023-09-22T17:40:34+08:00	warn	providers/controller.go:220	found a new leader apisix-internal-ingress-controller-54d9c8d775-9qwkf
2023-09-22T17:40:34+08:00	info	providers/controller.go:222	controller now is running as a candidate	{"namespace": "apisix", "pod": "apisix-internal-ingress-controller-5948cfdc95-hq8cm"}
2023-09-22T17:40:34+08:00	info	providers/controller.go:387	controller tries to leading ...	{"namespace": "apisix", "pod": "apisix-internal-ingress-controller-5948cfdc95-hq8cm"}
2023-09-22T17:40:34+08:00	warn	apisix/cluster.go:423	waiting cluster default to ready, it may takes a while
2023-09-22T17:40:34+08:00	info	apisix/cluster.go:248	syncing cache	{"cluster": "default"}
2023-09-22T17:40:34+08:00	info	apisix/cluster.go:463	syncing schema	{"cluster": "default"}
2023-09-22T17:40:34+08:00	error	apisix/route.go:90	failed to list routes: unexpected status code 503; error message: {"error_msg":"invalid response code: 502"}

2023-09-22T17:40:34+08:00	error	apisix/cluster.go:298	failed to list routes in APISIX: unexpected status code 503; error message: {"error_msg":"invalid response code: 502"}

2023-09-22T17:40:36+08:00	error	apisix/route.go:90	failed to list routes: unexpected status code 503; error message: {"error_msg":"invalid response code: 502"}

[...]

PR #2236, opened 2024-05-15
fix(leader-election): exit after Leader status is lost

Ben10k:
This PR is a subset of @acuteaura's PR #2152
The current problem (v1.8.2) is that whenever a controller loses it's `leader` status, it does not exit gracefully, thus it fails silently. In order to prevent this, an `os.exit` has been implemented to shut itself down, and depend on Kubernetes to bring it back up.

### Type of change:
- [x] Bugfix

ingress-controller doesn't recover from failed sync
#1980

### Pre-submission checklist:

### DIFF
--- a/pkg/providers/controller.go
+++ b/pkg/providers/controller.go
@@ -231,6 +231,9 @@ func (c *Controller) setupLeaderElection() error {
 					zap.String("namespace", c.namespace),
 					zap.String("pod", c.name),
 				)
+				// rootCancel might be to slow, and controllers may have bugs that cause them to not yield
+				// the safest way to step down is to simply cause a pod restart
+				os.Exit(0)
 			},
 		},
 		ReleaseOnCancel: true,
--- a/pkg/providers/controller.go
+++ b/pkg/providers/controller.go
@@ -185,6 +185,9 @@ func (c *Controller) Run(ctx context.Context) error {
 
 	c.run(rootCtx)
 
+	// c.run should never be returning when there's no error or the context is cancelled
+	rootCancel()
+
 	wg.Wait()
 	return nil
 }


acuteaura:
okay, sorry, i should've been more clear. the leader status is never lost, and that's the problem. the controller holds on to it indefinitely, even though it really shouldn't.

you can keep this bit of code and then call `rootCancel` after `run` is invoked [here](https://github.com/apache/apisix-ingress-controller/blob/d513e59a03c37c16fa62b44d50dbedb21b57e46a/pkg/providers/controller.go#L186), because that controller should never be returning when there's no error or the context is cancelled, and the context that keeps the leader election loop alive is derived from it.

Ben10k:
I have just pushed an update. @acuteaura can you take a look?

acuteaura:
Hm, I hadn't considered that the process also calls `run` and starts all the controller when it's not leader, so you're probably better off just hard-exiting instead of cancelling the context, because a follower node would never hit OnStoppedLeading. This will leave you with no leader until the lease expires, which is short enough (?!) so it shouldn't be a problem (it's essentially a "pod vanishes without trace" reenactment for the failover).

acuteaura:
@Revolyssup Would this arguably unclean but more reliable fix work for you?

wofr:
Would be great if this fix could be merged. We are running on GKE with out-of-date updates. After each update, we have a fifty-fifty chance of needing a manual restart of the ingress-controller.

bstasz:
Hello,
@acuteaura @Revolyssup What is missing or needed here to get this PR merged? Currently we are not able to deploy more than 1 instances as the Leader election will fail in 1-2 days.

Looking at the commit history, it feels like Apisix Ingress Controller is in maintenance mode. Is this a correct assessment or just deemed feature complete?

acuteaura:
this one's not correct. the correct way would be to revivie #2152 and fix the e2e test. or at least cherry-pick run to have an error return value and hard exit when it's not nil.

i wouldn't consider this project "dead" or "in maintenance mode", it's just very driven by individual contributors implementing what they need and some extra volunteers.

PR #2489, opened 2025-07-23
fix: full sync during restart results in loss of dataplane traffic

AlinsRan:
### Type of change:

- [x] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches
- [ ] Documentation
- [ ] Refactor
- [ ] Chore
- [ ] CI/CD or Tests

#### background

The controller watches CRDs and incrementally builds an in-memory cache of dataplane resources. It periodically performs a full (asynchronous) sync from this cache to the dataplane, without waiting for the cache to be fully populated.

On startup, if the number of CRDs is large, the first full sync may happen before all resources are processed, resulting in partial configuration being pushed to the dataplane. This can cause temporary 404 errors and traffic disruption.

#### Reproduction Steps:

1. Apply 5,000 ApisixRoute to the kubernetes cluster.

2. Restart the Ingress controller.

3. While the Ingress controller is starting, attempt to access all routes via the data plane. Some routes will return 404 Not Found errors.

### Pre-submission checklist:

### DIFF
--- a/internal/controller/apisixconsumer_controller.go
+++ b/internal/controller/apisixconsumer_controller.go
@@ -39,6 +39,7 @@ import (
 	apiv2 "github.com/apache/apisix-ingress-controller/api/v2"
 	"github.com/apache/apisix-ingress-controller/internal/controller/indexer"
 	"github.com/apache/apisix-ingress-controller/internal/controller/status"
+	"github.com/apache/apisix-ingress-controller/internal/manager/readiness"
 	"github.com/apache/apisix-ingress-controller/internal/provider"
 	"github.com/apache/apisix-ingress-controller/internal/utils"
 )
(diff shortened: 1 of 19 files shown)


internal/manager/controllers.go
```diff
@@ -176,3 +193,41 @@ func setupControllers(ctx context.Context, mgr manager.Manager, pro provider.Pro
...
+				types.GvkOf(&apiv2.ApisixGlobalRule{}),
+				types.GvkOf(&apiv2.ApisixPluginConfig{}),
+				types.GvkOf(&apiv2.ApisixTls{}),
+				types.GvkOf(&apiv2.ApisixConsumer{}),
+			},
+			Filter: readiness.GVKFilter(func(obj *unstructured.Unstructured) bool {
+				icName, _, _ := unstructured.NestedString(obj.Object, "spec", "ingressClassName")
+				ingressClass, _ := controller.GetIngressClass(context.Background(), c, log, icName)
+				return ingressClass != nil
+			}),
```
ronething, line 218: If there is no ingressClassName field, will we ignore it directly?

AlinsRan: The `GetIngressClass` has processing logic for `ingressClassName` to be empty.

The `ingressClass` object will only be ignored if it does not exist

internal/controller/apisixconsumer_controller.go
```diff
@@ -51,11 +52,13 @@ type ApisixConsumerReconciler struct {
 
 	Provider provider.Provider
 	Updater  status.Updater
+	Readier  readiness.ReadinessManager
 }
 
 // Reconcile FIXME: implement the reconcile logic (For now, it dose nothing other than directly accepting)
 func (r *ApisixConsumerReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	r.Log.Info("reconcile", "request", req.NamespacedName)
+	r.Readier.Done(&apiv2.ApisixConsumer{}, req.NamespacedName)
```
ronething: It would be more reasonable to execute the packaging within `defer func`

nic-6443 (approved):

PR #2495, opened 2025-07-29
fix: traffic-split weight distribution and add e2e tests

Revolyssup:
This PR solves two bugs and adds corresponding E2E tests:
1. `backendErr` is currently a shared variable due to which even if one backend is invalid, no backend is set causing 0 nodes to be set on upstream even when there are valid nodes. A test case confirming the fix for this has been added. The test `valid backend is set even if other backend is invalid` confirms this fix.
2. The weight from backend was not set in the upstream  label which is used to set weight on `weightedUpstreams` due to which all nodes always get `100` as weight. The tests `2:1 traffic split test` and `zero-weight test` confirm this fix.
- [x] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches
- [ ] Documentation
- [ ] Refactor
- [ ] Chore
- [x] CI/CD or Tests

### Pre-submission checklist:

### DIFF
--- a/test/e2e/crds/v2/route.go
+++ b/test/e2e/crds/v2/route.go
@@ -744,7 +744,6 @@ spec:
 				Expect(code).Should(Equal(200))
 			}
 		})
-
 		It("valid backend is set even if other backend is invalid", func() {
 			const apisixRouteSpec = `
 apiVersion: apisix.apache.org/v2
@@ -781,6 +780,8 @@ spec:
 				code := verifyRequest()
 				Expect(code).Should(Equal(200))
 			}
+		})
+	})
 	Context("Test ApisixRoute sync during startup", func() {
 		const route = `
 apiVersion: apisix.apache.org/v2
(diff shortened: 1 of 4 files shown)


test/e2e/crds/v2/route.go
```diff
@@ -651,6 +652,136 @@ spec:
...
+			verifyRequest := func() int {
+				return s.NewAPISIXClient().GET("/get").WithHost("httpbin.org").Expect().Raw().StatusCode
+			}
+			By("send requests to verify traffic split")
+			var (
+				successCount int
+				failCount    int
+			)
+
+			time.Sleep(10 * time.Second)
```
AlinsRan: We can use request detection to avoid prolonged sleep.
ref: https://github.com/apache/apisix-ingress-controller/blob/master/test/e2e/gatewayapi/httproute.go#L351-L356

AlinsRan (approved):

PR #2501, opened 2025-07-31
refactor:  E2E tests to support parallel tests

Revolyssup:
- [ ] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches
- [ ] Documentation
- [x] Refactor
- [ ] Chore
- [x] CI/CD or Tests

Currently running tests in parallel doesn't work due to conflict of Shared resources. For instance, the tests in crds/v2 share GatewayProxy and IngressClass configuration causing issues while running tests in parallel. 
Goal of this PR is to eliminate the use of shared resources and use unique resources for tests.

Eventually from here and now, we need unique resource <name/namespace> for GatewayProxy, IngressClass, Gateway, GatewayClass, Ingress(similarly HTTPRoute, etc) so that parallel tests don't override each other's configuration. 
In this PR, I have replaced all static names with `s.Namespace()` as it already provides a unique string.

### Things to note before review:
- In the test helper function, if replica is 0 then test will not tunnel and wait for pod.
- The test that require to share resource like "IngressClass" for eg: testing default ingressclass cannot be run in parallel therefore have been marked as  `Serial`.
- To make tests more resilient, direct request calls have been replaced with Eventully() block to retry until success avoiding Sleep.
- The timeout has been increased for retries.

### Pre-submission checklist:


Revolyssup:
Another way of doing this is to create a random name generator like in this PR
: https://github.com/Revolyssup/apisix-ingress-controller/pull/4/files#diff-d32745cfa87775abccdb567bf340d4532ae85ca97ed745d25bd56337e2c08658

If this method looks better, then reviewers can suggest.

AlinsRan:
https://github.com/apache/apisix-ingress-controller/actions/runs/16802790400/job/47587870938?pr=2501#step:10:16666

Each job ran all the use cases.
 Referring to the previous [e2e-test](https://github.com/api7/api7-ingress-controller/blob/release-v2-dev/Makefile#L138), we need to add label-filter

test/e2e/ingress/ingress.go
```diff
@@ -231,81 +232,93 @@ spec:
...
-				Expect().
-				Status(http.StatusOK)
-
-			s.NewAPISIXClient().
-				GET("/get").
-				WithHost("default.example.com").
-				Expect().
-				Status(200)
-
+			time.Sleep(10 * time.Minute)
```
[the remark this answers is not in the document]
Revolyssup: wow I really forgot to remove this😂

.github/workflows/apisix-e2e-test.yml
```diff
@@ -73,6 +73,12 @@ jobs:
         with:
           go-version: "1.23"
 
+      - name: Install ginkgo
+        run: |
+            GINKGO_VERSION=$(cd test/e2e && go list -m  -mod=readonly  -f {{.Version}}  github.com/onsi/ginkgo/v2)
+            go install github.com/onsi/ginkgo/v2/ginkgo@$GINKGO_VERSION
```
ronething: use make install-ginkgo?

test/e2e/crds/v1alpha1/backendtrafficpolicy.go
```diff
@@ -19,21 +19,28 @@ package v1alpha1
...
+	"k8s.io/apimachinery/pkg/types"
 
 	"github.com/apache/apisix-ingress-controller/test/e2e/scaffold"
 )
 
 var _ = Describe("Test BackendTrafficPolicy base on HTTPRoute", Label("apisix.apache.org", "v1alpha1", "backendtrafficpolicy"), func() {
-	s := scaffold.NewDefaultScaffold()
+	var (
+		s = scaffold.NewScaffold(&scaffold.Options{
+			ControllerName: fmt.Sprintf("apisix.apache.org/apisix-ingress-controller-%d", time.Now().Unix()),
```
ronething, line 34: This can be used as the default behavior.

AlinsRan: It already has default behavior:
https://github.com/apache/apisix-ingress-controller/blob/master/test/e2e/scaffold/apisix_deployer.go#L69

test/e2e/crds/v1alpha1/backendtrafficpolicy.go
```diff
@@ -125,7 +133,26 @@ spec:
...
+			time.Sleep(time.Second)
+
+			By("create GatewayClass")
+			gatewayClassName := fmt.Sprintf("apisix-%d", time.Now().Unix())
+			err = s.CreateResourceFromString(fmt.Sprintf(defaultGatewayClass, gatewayClassName, s.GetControllerName()))
+			Expect(err).NotTo(HaveOccurred(), "creating GatewayClass")
+			time.Sleep(time.Second)
+
+			By("create Gateway")
+			err = s.CreateResourceFromStringWithNamespace(fmt.Sprintf(defaultGateway, gatewayName, gatewayClassName, gatewayProxyName), s.Namespace())
```
AlinsRan: `CreateResourceFromString` uses the namespace of the current test, so there's no need to explicitly specify it for namespace-scoped resources.

test/e2e/crds/v2/config.go
```diff
@@ -0,0 +1,61 @@
...
+    apiGroup: "apisix.apache.org"
+    kind: "GatewayProxy"
+    name: "%s"
+    namespace: "%s"
+    scope: "Namespace"
+`
+
+func getGatewayProxyYaml(namespace, endpoint, adminKey string) string {
+	return fmt.Sprintf(gatewayProxyYaml, namespace, namespace, endpoint, adminKey)
+}
```
ronething: The function can be encapsulated in the `Scaffold`, so we don't need to manually pass many parameters.

test/e2e/crds/v2/route.go
```diff
@@ -42,20 +42,20 @@ import (
...
-		gatewayProxy := fmt.Sprintf(gatewayProxyYaml, s.Deployer.GetAdminEndpoint(), s.AdminKey())
-		err := s.CreateResourceFromStringWithNamespace(gatewayProxy, "default")
+		gatewayProxy := getGatewayProxyYaml(s.Namespace(), s.Deployer.GetAdminEndpoint(), s.AdminKey())
+		err := s.CreateResourceFromStringWithNamespace(gatewayProxy, s.Namespace())
 		Expect(err).NotTo(HaveOccurred(), "creating GatewayProxy")
 		time.Sleep(5 * time.Second)
 
 		By("create IngressClass")
-		err = s.CreateResourceFromStringWithNamespace(ingressClassYaml, "")
+		err = s.CreateResourceFromStringWithNamespace(getIngressClassYaml(s.Namespace(), s.GetControllerName(), s.Namespace()), "")
```
ronething: for ingressClass，we only need to use `CreateResourceFromString`

test/e2e/crds/v1alpha1/backendtrafficpolicy.go
AlinsRan: Only cluster level resources will use `CreateResourceFromStringWithnamespace`, or a fixed namespace.

test/e2e/crds/v1alpha1/backendtrafficpolicy.go
Revolyssup: revered to defaultscaffold

test/e2e/crds/v2/route.go
Revolyssup: okay

test/e2e/crds/v2/config.go
Revolyssup: okay

test/e2e/crds/v1alpha1/backendtrafficpolicy.go
Revolyssup: I have reverted back to this behavior. I see a lot of tests fail when defaultScaffold is used. I am looking into the reason but for now I have reverted to this.

test/e2e/crds/v2/route.go
```diff
@@ -68,8 +68,9 @@ apiVersion: apisix.apache.org/v2
 kind: ApisixRoute
 metadata:
   name: default
+  namespace: %s
```
AlinsRan, line 71: For 95% of use cases, it is always applied in the namespace where the current use case is located. Do we still need to specify the namespace?

Revolyssup: Yes maybe its redundant, do you want to me to remove it from everywhere? It doesn't matter except for maybe these changes might have added strain in the review process. For a good practice, I was being explicit with namespaces so that in the rest 5% of the cases I dont make a mistake if I make a general rule to specify namespaces explicitly.

AlinsRan: Well, I don't think so. On the contrary, I think it will bring greater usage burden, but it's just a small issue.

PR #2507, opened 2025-08-13
chore: refactor provider

AlinsRan:
### Type of change:

- [ ] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches
- [ ] Documentation
- [ ] Refactor
- [x] Chore
- [ ] CI/CD or Tests

#### Description
This commit refactors the `adc` provider module with the goal of optimizing package structure, reducing module coupling, and improving code maintainability and extensibility.

##### Changes include
1. **Directory and package structure reorganization**
   - Moved code from the original `internal/provider/adc` directory to a new `internal/adc` directory.
   - Split the code into sub-packages such as `cache`, `client`, and `translator`, with clearer responsibilities.
   - Split configuration management into a `configmanger` component
2. **Decoupling and responsibility separation**
   - Extracted caching, execution, and resource translation logic from the provider core into separate modules to reduce internal dependencies.

##### Reason for the changes
- In the old structure, the `adc` provider’s various functionalities were tightly coupled at the same level, making maintenance and extension costly.
- With modular separation, it becomes easier to perform unit testing, replace implementations, or introduce new functional modules in the future.

### Pre-submission checklist:


juzhiyuan:
Just a reminder: when marking this PR as ready for review, please fill in the description and checklist completely.

.github/workflows/apisix-conformance-test.yml
```diff
@@ -89,7 +89,7 @@ jobs:
 
       - name: Install And Run Cloud Provider KIND
         run: |
-          go install sigs.k8s.io/cloud-provider-kind@latest
+          go install sigs.k8s.io/cloud-provider-kind@v0.6.0
```
bzp2010, line 92: The latest version is `0.7.0`, why can't we use `latest`?

api/adc/types.go
bzp2010, line 1: The `api` folder seems to be used to store CRDs. Would it be appropriate to place an ADC type here?

internal/provider/common/configmanager.go
```diff
@@ -0,0 +1,146 @@
...
+// specific language governing permissions and limitations
+// under the License.
+
+package common
+
+import (
+	"sync"
+)
+
+type ConfigManager[K comparable, T any] struct {
```
bzp2010, line 46: Please add some comments to point out the purpose of this structure.

.github/workflows/apisix-conformance-test.yml
AlinsRan: https://github.com/apache/apisix-ingress-controller/actions/runs/16980187972/job/48138378106

Go.mod uses 1.23

api/adc/types.go
AlinsRan: The definition of CRDs is not constrained and depends on the project itself

internal/provider/common/configmanager.go
AlinsRan: Please check it.

.github/workflows/apisix-conformance-test.yml
bzp2010: Based on what go releases have done, it doesn't seem to make sense for us to be stuck with older versions, newer versions always include CVE security fixes and runtime improvements.

Maybe we should upgrade it in a later PR.

PR #2528, opened 2025-08-26
fix: hmac-auth plugin spec compatibility with latest apisix

Revolyssup:
The hmac-auth plugin in APISIX had gone through a rewrite/refactor due to which it's incompatible with current version of ingress controller. This PR makes it compatible with https://github.com/apache/apisix/pull/11581

Key changes:
- Other than `key_id` and `secret_key` all fields removed from consumer configuration.
- `access_key` renamed to `key_id`
- [ ] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches
- [ ] Documentation
- [ ] Refactor
- [ ] Chore
- [ ] CI/CD or Tests

### Pre-submission checklist:


AlinsRan (changes requested):
This change introduces a breaking change. 
When using older versions of the data plane, this API still applies.
Instead of removing the existing field directly, we should mark it as deprecated and clearly document the incompatibility, so that users have a migration path. Meanwhile, we can introduce the new field to support the latest data plane.

AlinsRan (approved):

PR #2534, opened 2025-09-01
feat: support retry in case of sync failure

AlinsRan:
### Type of change:

- [ ] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches
- [ ] Documentation
- [ ] Refactor
- [ ] Chore
- [ ] CI/CD or Tests

### Pre-submission checklist:

### DIFF
--- a/internal/provider/apisix/provider.go
+++ b/internal/provider/apisix/provider.go
@@ -37,6 +37,7 @@ import (
 	"github.com/apache/apisix-ingress-controller/internal/controller/status"
 	"github.com/apache/apisix-ingress-controller/internal/manager/readiness"
 	"github.com/apache/apisix-ingress-controller/internal/provider"
+	"github.com/apache/apisix-ingress-controller/internal/provider/common"
 	"github.com/apache/apisix-ingress-controller/internal/types"
 	"github.com/apache/apisix-ingress-controller/internal/utils"
 )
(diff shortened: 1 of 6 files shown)


internal/provider/apisix/provider.go
```diff
@@ -223,32 +231,38 @@ func (d *apisixProvider) Start(ctx context.Context) error {
...
-		})
+		time.AfterFunc(initalSyncDelay, d.syncNotify)
 	}
 
-	if d.SyncPeriod < 1 {
-		return nil
+	syncPeriod := d.SyncPeriod
+	if syncPeriod < MinSyncPeriod {
+		syncPeriod = MinSyncPeriod
 	}
```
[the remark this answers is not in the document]
AlinsRan: https://github.com/apache/apisix-ingress-controller/blob/master/internal/controller/config/config.go#L122

internal/provider/common/retrier.go
```diff
@@ -0,0 +1,91 @@
...
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	delay := r.backoff.Next()
+	r.timer = time.AfterFunc(delay, func() {
+		select {
+		case r.ch <- struct{}{}:
+		default:
+		}
+	})
```
ronething, line 91: If the Trigger is triggered multiple times, and it is not necessarily from the last timer trigger (for example, triggered by the syncPeriod synchronization time and also failed to synchronize), does this indicate a timer leak here?

AlinsRan: If it keeps failing, it will continue to run until it succeeds. At the same time, it will also be reset by other events. It will not run again after successful synchronization.

AlinsRan: You're right, this will burden the user. It's more appropriate to simply stop the timer and proceed to the next one.

Revolyssup (approved):

PR #2538, opened 2025-09-02
fix: sync exception caused by ingress endpoint 0

AlinsRan:
### Type of change:

- [ ] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches
- [ ] Documentation
- [ ] Refactor
- [ ] Chore
- [ ] CI/CD or Tests

### Pre-submission checklist:

### DIFF
--- a/internal/adc/translator/ingress.go
+++ b/internal/adc/translator/ingress.go
@@ -222,7 +222,7 @@ func (t *Translator) TranslateIngress(tctx *provider.TranslateContext, obj *netw
 
 // translateEndpointSliceForIngress create upstream nodes from EndpointSlice
 func (t *Translator) translateEndpointSliceForIngress(weight int, endpointSlices []discoveryv1.EndpointSlice, servicePort *corev1.ServicePort) adctypes.UpstreamNodes {
-	var nodes adctypes.UpstreamNodes
+	nodes := adctypes.UpstreamNodes{}
 	if len(endpointSlices) == 0 {
 		return nodes
 	}
--- a/test/e2e/ingress/ingress.go
+++ b/test/e2e/ingress/ingress.go
@@ -204,8 +204,7 @@ spec:
             port:
               number: 80
 `
-
-		It("Test IngressClass Selection", func() {
+		BeforeEach(func() {
 			By("create GatewayProxy")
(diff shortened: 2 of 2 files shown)


api/adc/types.go
```diff
@@ -505,6 +505,20 @@ func (n *UpstreamNodes) UnmarshalJSON(p []byte) error {
...
+// MarshalJSON implements the json.Marshaler interface for UpstreamNodes.
+// By default, Go serializes a nil slice as JSON null. However, for compatibility
+// with APISIX semantics, we want a nil UpstreamNodes to be encoded as an empty
+// array ([]) instead of null. Non-nil slices are marshaled as usual.
+//
+// See APISIX upstream nodes schema definition for details:
+// https://github.com/apache/apisix/blob/77dacda31277a31d6014b4970e36bae2a5c30907/apisix/schema_def.lua#L295-L338
+func (n UpstreamNodes) MarshalJSON() ([]byte, error) {
+	if n == nil {
+		return []byte("[]"), nil
```
bzp2010, line 517: Isn't this achieved by filling empty slices as fallback values? This approach doesn't seem conducive to maintenance.
An overridden implementation would require digging deep into the code to discover.

That is, it is implemented during translation rather than during encoding.

https://go.dev/play/p/rB6BCY0n670

AlinsRan: * UpstreamNodes is part of the external API contract; the schema requires it to always be an array ([]), never null.
* Overriding MarshalJSON ensures consistent output regardless of whether the slice is initialized.
* This is defensive programming, preventing new code or omissions from causing the API to return null and break the contract.
* Compared to relying on every construction point to initialize the slice, this approach centralizes the handling and reduces maintenance overhead.
* The comment clearly explains the purpose and includes a schema link for reference.

https://go.dev/play/p/lS8wac0gRQZ

bzp2010 (approved):

api/adc/types.go
bzp2010: Perhaps it's best to do this during both translation and encoding.

The implementation of the Marshaler interface is implicit; without examining the function implementations on UpstreamNodes, one cannot understand its specific operations. This requires familiarity with the codebase to explain why this data structure diverges from standard Go JSON encoding behavior.
I agree this should serve as a fallback, but only as a fallback. This doesn't mean we should overlook potential errors in the translation process.

For suggestions only.

PR #2543, opened 2025-09-04
fix: residual data issue when updating ingressClassName

AlinsRan:
### Type of change:

- [x] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches
- [ ] Documentation
- [ ] Refactor
- [ ] Chore
- [ ] CI/CD or Tests

In the current implementation, the controller attempts to retrieve the IngressClass associated with an Ingress resource that matches the specified controllerName. If the IngressClass cannot be retrieved, the controller simply ignores the Ingress resource. However, this behavior is problematic: the corresponding configuration is not removed and remains in the data plane, which may result in incorrect routing or stale configurations.

Therefore, when the ingressClassName changes, the old configuration must be properly cleaned up and removed to ensure consistency between the control plane and the data plane, and to prevent leftover or conflicting configurations.

#### Reproduce steps:

1. apply `ApisixRoute`
```yaml
kind: ApisixRoute
metadata:
  name: default
spec:
  ingressClassName: apisix
  http:
  - name: rule0
    match:
      hosts:
      - httpbin
      paths:
      - /get
    backends:
    - serviceName: httpbin
      servicePort: 80
```

2. Access route successful

```shell
$ curl http://localhost:9080/get -H "Host: httpbin"
200
```

3. update invalid ingressClassName

```yaml
kind: ApisixRoute
metadata:
  name: default
spec:
  ingressClassName: invalid
  http:
  - name: rule0
    match:
      hosts:
      - httpbin
      paths:
      - /get
    backends:
    - serviceName: httpbin
      servicePort: 80
```

4. Expected to receive a response status code of 404, but actually received 200

```shell
$ curl http://localhost:9080/get -H "Host: httpbin"
200
```

### Pre-submission checklist:

### DIFF
--- a/test/e2e/framework/manifests/ingress.yaml
+++ b/test/e2e/framework/manifests/ingress.yaml
@@ -326,6 +326,15 @@ data:
     log_level: "debug"
     controller_name: {{ .ControllerName | default "apisix.apache.org/apisix-ingress-controller" }}
     leader_election_id: "apisix-ingress-controller-leader"
+    leader_election:
+      lease_duration: 10s                   # lease_duration is the duration that non-leader candidates will wait
+                                            # after observing a leadership renewal until attempting to acquire leadership of a
+                                            # leader election.
+      renew_deadline: 5s                   # renew_deadline is the time in seconds that the acting controller
+                                            # will retry refreshing leadership before giving up.
+      retry_period: 2s                      # retry_period is the time in seconds that the acting controller
+                                            # will wait between tries of actions with the controller.
+      disable: false                        # Whether to disable leader election.
     exec_adc_timeout: 5s
     provider:
       type: {{ .ProviderType | default "apisix" }}


internal/controller/apisixglobalrule_controller.go
```diff
@@ -84,11 +82,15 @@ func (r *ApisixGlobalRuleReconciler) Reconcile(ctx context.Context, req ctrl.Req
 	tctx := provider.NewDefaultTranslateContext(ctx)
 
 	// get the ingress class
-	ingressClass, err := GetIngressClass(tctx, r.Client, r.Log, globalRule.Spec.IngressClassName)
+	ingressClass, err := FindMatchingIngressClass(tctx, r.Client, r.Log, &globalRule)
 	if err != nil {
```
bzp2010, line 86: If there is a request error caused by the unavailability of apiserver, will it delete the resource unexpectedly?

I see that there is no special treatment for this in your matching function.

internal/controller/utils.go
```diff
@@ -1368,7 +1339,12 @@ func ProcessIngressClassParameters(tctx *provider.TranslateContext, c client.Cli
 	return nil
 }
 
-func GetIngressClass(ctx context.Context, c client.Client, log logr.Logger, ingressClassName string) (*networkingv1.IngressClass, error) {
+func FindMatchingIngressClass(ctx context.Context, c client.Client, log logr.Logger, obj client.Object) (*networkingv1.IngressClass, error) {
+	ingressClassName := ExtractIngressClass(obj)
+	return FindMatchingIngressClassByName(ctx, c, log, ingressClassName)
+}
+
+func FindMatchingIngressClassByName(ctx context.Context, c client.Client, log logr.Logger, ingressClassName string) (*networkingv1.IngressClass, error) {
```
bzp2010, line 1347: This function does not handle the request failure caused by network failure. This will cause unintented delete behavior.

internal/controller/apisixglobalrule_controller.go
AlinsRan: https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/client/client.go#L343-L362

The behavior here is to query the local cache.
 IngressClass is our main watch resource, so there won't be any network issues.

AlinsRan: The same goes for `list`, and it is expected that read operations on resources will not generate network calls.

internal/controller/utils.go
AlinsRan: IngressClass is the main watch object, and all behavior is based on local caching, so it does not generate any network calls.

bzp2010 (approved):

PR #2550, opened 2025-09-09
feat: add unified API server with debugging capabilities

Revolyssup:
Goals:
- Provide an API endpoint for debugging.
- Provide an API that generates translated ADC configurations based on the latest Gateway API/Ingress/CRD resources for inspection and debugging, outputting them in JSON.
- Provide a homepage guide. Use pure HTML to create a simple title and link list.

Solution:
- Uses the underlying store that adc uses to sync data as the store for debug handler.
- Create a new runnable `Server` that registers handlers on it and runs the server.

https://github.com/user-attachments/assets/550ad845-417f-4a05-b66e-a4645350aabe

- [ ] Bugfix
- [x] New feature provided
- [ ] Improve performance
- [ ] Backport patches
- [ ] Documentation
- [ ] Refactor
- [ ] Chore
- [ ] CI/CD or Tests

### Pre-submission checklist:

### DIFF
--- a/internal/provider/common/adcdebugserver.go
+++ b/internal/provider/common/adcdebugserver.go
@@ -28,15 +28,6 @@ import (
 	"github.com/apache/apisix-ingress-controller/internal/types"
 )
 
-const (
-	consumers      = "consumers"
-	services       = "services"
-	routes         = "routes"
-	ssls           = "ssls"
-	globalrules    = "globalrules"
-	pluginmetadata = "pluginmetadata"
-)
-
 type ResourceInfo struct {
 	ID   string
 	Name string
@@ -132,7 +123,7 @@ func (asrv *ADCDebugProvider) handleConfig(w http.ResponseWriter, r *http.Reques
 }
 
 func (asrv *ADCDebugProvider) showResourceTypes(w http.ResponseWriter, configName, configNameEncoded string) {
-	resourceTypes := []string{"services", "routes", "consumers", "ssls", "globalrules", "pluginmetadata"}
+	resourceTypes := []string{adctypes.TypeService, adctypes.TypeRoute, adctypes.TypeConsumer, adctypes.TypeSSL, adctypes.TypeGlobalRule, adctypes.TypePluginMetadata}
 
 	tmpl := newTemplate("resources", `
         <html>
@@ -170,7 +161,7 @@ func (asrv *ADCDebugProvider) showResources(w http.ResponseWriter, r *http.Reque
 
 	var resourceInfos []ResourceInfo
 	switch resourceType {
-	case services:
+	case adctypes.TypeService:
 		for _, svc := range resources.Services {
 			resourceInfos = append(resourceInfos, ResourceInfo{
 				ID:   svc.ID,
@@ -180,7 +171,7 @@ func (asrv *ADCDebugProvider) showResources(w http.ResponseWriter, r *http.Reque
 					asrv.pathPrefix, configNameEncoded, url.QueryEscape(resourceType), url.QueryEscape(svc.ID)),
 			})
 		}
-	case consumers:
+	case adctypes.TypeConsumer:
 		for _, consumer := range resources.Consumers {
 			resourceInfos = append(resourceInfos, ResourceInfo{
 				ID:   consumer.Username,
@@ -190,7 +181,7 @@ func (asrv *ADCDebugProvider) showResources(w http.ResponseWriter, r *http.Reque
 					asrv.pathPrefix, configNameEncoded, url.QueryEscape(resourceType), url.QueryEscape(consumer.Username)),
 			})
 		}
-	case ssls:
+	case adctypes.TypeSSL:
 		for _, ssl := range resources.SSLs {
 			resourceInfos = append(resourceInfos, ResourceInfo{
 				ID:   ssl.ID,
@@ -200,7 +191,7 @@ func (asrv *ADCDebugProvider) showResources(w http.ResponseWriter, r *http.Reque
 					asrv.pathPrefix, configNameEncoded, url.QueryEscape(resourceType), url.QueryEscape(ssl.ID)),
 			})
 		}
-	case globalrules:
+	case adctypes.TypeGlobalRule:
 		for key := range resources.GlobalRules {
 			resourceInfos = append(resourceInfos, ResourceInfo{
 				ID:   key,
@@ -210,7 +201,7 @@ func (asrv *ADCDebugProvider) showResources(w http.ResponseWriter, r *http.Reque
 					asrv.pathPrefix, configNameEncoded, url.QueryEscape(resourceType), url.QueryEscape(key)),
 			})
 		}
-	case pluginmetadata:
+	case adctypes.TypePluginMetadata:
 		if resources.PluginMetadata != nil {
 			resourceInfos = append(resourceInfos, ResourceInfo{
 				ID:   "pluginmetadata",
@@ -220,7 +211,7 @@ func (asrv *ADCDebugProvider) showResources(w http.ResponseWriter, r *http.Reque
 					asrv.pathPrefix, configNameEncoded, url.QueryEscape(resourceType), "pluginmetadata"),
 			})
 		}
-	case routes:
+	case adctypes.TypeRoute:
 		for _, svc := range resources.Services {
 			for _, route := range svc.Routes {
 				resourceInfos = append(resourceInfos, ResourceInfo{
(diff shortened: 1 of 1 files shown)


internal/provider/apisix/provider.go
```diff
@@ -80,12 +81,14 @@ func New(updater status.Updater, readier readiness.ReadinessManager, opts ...pro
...
 	return &apisixProvider{
-		client:     cli,
-		Options:    o,
-		translator: &translator.Translator{},
-		updater:    updater,
-		readier:    readier,
-		syncCh:     make(chan struct{}, 1),
+		client: cli,
+		// TODO: Maybe pass port/address from external configuration
+		adcdebugserver: common.NewADCDebugServer(cli.Store, cli.ConfigManager, 8432),
```
Revolyssup: I don't think it should be needed to make this port configurable. Maybe we can hardcode it to some value.

AlinsRan:
I don't think it should be a standalone server, rather than a part of the ingress API.

internal/manager/server/server.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+func (s *Server) Register(pathPrefix string, registrant provider.RegisterHandler) {
+	subMux := http.NewServeMux()
+	registrant.Register(pathPrefix, subMux)
+	s.mux.Handle(pathPrefix+"/", http.StripPrefix(pathPrefix, subMux))
+	s.mux.HandleFunc(pathPrefix, func(w http.ResponseWriter, r *http.Request) {
+		http.Redirect(w, r, pathPrefix+"/", http.StatusPermanentRedirect)
+	})
+}
```
ronething, line 58: The debug server currently has no authentication mechanism, making it accessible to anyone who can reach the endpoint. Should we consider incorporating certain authentication mechanisms?

Revolyssup: The original requirement was to return simple UI friendly HTML. Adding authentication will mean adding another page for login. Maybe we can use cookie based authentication and add one more simple login html page.  
1. Check if cookie has `INGRESS-DEBUG_TOKEN`. If present then serve normally.
2. If not present then redirect to login page. This will also be a simple HTML page with just input box to enter `debug token`. This `debug token` will be set in the static configuration for ingress and can be copied from there and pasted here. User clicks `login` and this value be just stored in the cookie if matches and user will be logged in.

@bzp2010  @ronething @AlinsRan What do you think?

AlinsRan: I think it's enough not to expose the port, just like the control-api.

internal/provider/apisix/provider.go
```diff
@@ -80,15 +82,20 @@ func New(updater status.Updater, readier readiness.ReadinessManager, opts ...pro
...
 
 	return &apisixProvider{
-		client:     cli,
-		Options:    o,
-		translator: &translator.Translator{},
-		updater:    updater,
-		readier:    readier,
-		syncCh:     make(chan struct{}, 1),
+		client:           cli,
+		adcDebugProvider: common.NewADCDebugProvider(cli.Store, cli.ConfigManager),
```
AlinsRan: Why does it belong to the provider rather than the ADC client?

Revolyssup: Makes sense. move it.

internal/manager/server/server.go
ronething: > I think it's enough not to expose the port, just like the control-api.

It might be a bit different. This debug server has exposed SSL/Consumer resources, but the control-api does not.

AlinsRan: > It might be a bit different. This debug server has exposed SSL/Consumer resources, but the control-api does not.

You're right.

bzp2010: From a security standpoint, listening on 127.0.0.1 is the minimum requirement. We could disable this feature by default and enable it selectively via configuration files. If we do this, would we need some mechanism to safely enable this feature dynamically without requiring a restart?

Alternatively, if we can guarantee that the ingress controller's output remains fixed for a given input regardless of how it restarts, we could configure it to enable this feature only after a restart.

@AlinsRan @ronething @Revolyssup

Revolyssup: I think the only k8s native way to enable/disable this dynamically will be to use some existing CR for it. But that might be an overkill. I think we should opt for one of the two ways:
1. Just disable this server by default and have it enabled statically only and not dynamically. It's a debug server, I don't know if dynamically enabling/disabling it is good enough use case. This is the simplest and safest. 
2. Just have a single login page that expects a statically defined key as recommended above.

If it's okay to enable/disable this statically then I recommend the first way is good enough or else it will be overengineering. Later if use case emerges about dynamically enabling/disabling, we can spend some thought here.

bzp2010: Yes, that's exactly what I was thinking. A simple static config.

However, I must confirm whether the restart required to enable this feature would disrupt the current "state". Specifically, can we guarantee consistent results based on unchanged configurations? We need to ensure that the restart does not corrupt any potentially existing in-memory state, thereby preventing the debugging of previously existing issues.

For example, we sometimes encounter problems that resolve after a restart, making it impossible to analyze the root cause.

Can anyone answer this question?

Revolyssup: The restart will rebuild the in-memory store which is the data store for this API. Given same set of configuration, theoretically it should go rebuild the same state according to me. Because the data from this in-mem store directly corresponds to data that will be sent to data plane, in case there are inconsistencies on restart this has to be a bug.

If there are problems that resolve after restart, then we should note them and debug to find the root cause. The expectation from the system should be such that it produces consistently same state for the same input. And on this expectation of the system we can go with the static config approach. I cannot think of a case where it will be expected from the system to produce different state on reruns.

AlinsRan: It’s not possible to guarantee the preservation of “state” across a restart.  
The controller’s declarative configuration state (stored in Kubernetes/etcd via CRDs) is automatically rebuilt after a restart, but any runtime in-memory state (caches, queues, goroutine stacks, etc.) will inevitably be lost.  

For an ingress controller, a restart is typically the last-resort mechanism to recover state, so if you need to debug live in-memory data, it will naturally no longer be available after restarting.  

At present, for security and simplicity reasons, we recommend keeping the debug API disabled by default and enabling it only via static configuration.  
If there is a strong need to inspect runtime state in the future, we can revisit supporting dynamic enablement or enabling it up front with proper authentication.

Revolyssup: done. added the boolean in static config.

bzp2010: The state itself isn't critical, but it would be preferable if you could confirm it doesn't impact configuration generation.

The key point is `For a given input of Kubernetes resources, always produce a fixed output.` This includes how to handle conflicting resources (determine the priority), how to sort resource lists, and so on. This process already exists within AIC, such as during full synchronization at startup. We can ignore those changes triggered by reconciliation tasks.
If these could be leveraged for debugging functionality, that would be excellent.

config/samples/config.yaml
```diff
@@ -18,6 +18,8 @@ leader_election:
 
 metrics_addr: ":8080"                   # The address the metrics endpoint binds to.
                                         # The default value is ":8080".
+enable_server: false                    # The debug API is behind this server which is disabled by default for security reasons.
+server_addr: ":9092"
```
AlinsRan: 127.0.0.1?
Should we emphasize the purpose of debugging on the fields?

Revolyssup: binded to localhost.

Revolyssup: The purpose is generic. Although currently it's to debug the in memory state of adc config. Maybe we can add a list in the comment.
1. `/debug` can be used to debug in-memory state of translated adc configs to be synced with data plane.
....later if handlers are added, we can extend the list

Revolyssup: added

internal/provider/common/adcdebugserver.go
```diff
@@ -0,0 +1,341 @@
...
+		for _, svc := range resources.Services {
+			resourceInfos = append(resourceInfos, ResourceInfo{
+				ID:   svc.ID,
+				Name: svc.Name,
+				Type: resourceType,
+				Link: fmt.Sprintf("%s/config?name=%s&type=%s&id=%s",
+					asrv.pathPrefix, configNameEncoded, url.QueryEscape(resourceType), url.QueryEscape(svc.ID)),
+			})
+		}
+	case "consumers":
```
AlinsRan: Avoid using hard coding.

Revolyssup: done

AlinsRan: https://github.com/apache/apisix-ingress-controller/blob/4e1bd6eb43591ae28c541bd43567955a02f6f829/api/adc/types.go#L80-L88

Revolyssup: replaced with adctypes

config/samples/config.yaml
```diff
@@ -18,6 +18,8 @@ leader_election:
 
 metrics_addr: ":8080"                   # The address the metrics endpoint binds to.
                                         # The default value is ":8080".
+enable_server: false                    # The debug API is behind this server which is disabled by default for security reasons.
```
bzp2010, line 21: The `server` may be ambiguous. For example, we also have a metrics server (and `metrics_addr`). Does this need any tuning?

internal/provider/common/adcdebugserver.go
```diff
@@ -0,0 +1,341 @@
...
+func newTemplate(name, body string) *template.Template {
+	return template.Must(template.New(name).
+		Funcs(template.FuncMap{"urlencode": url.QueryEscape}).
+		Parse(body))
+}
+
+func (asrv *ADCDebugProvider) SetupHandler(pathPrefix string, mux *http.ServeMux) {
+	asrv.pathPrefix = pathPrefix
+	mux.HandleFunc("/config", asrv.handleConfig)
+	mux.HandleFunc("/", asrv.handleIndex)
```
bzp2010, line 53: Can those APIs ensure that the resources they output are managed by the current controller? For example, if an HTTPRoute is managed by another ingress controller, can it be listed by the current API?

config/samples/config.yaml
Revolyssup: I couldn't think of a better name because it's a generic server and currently has debugging related endpoints but can be extended. Do you have a better suggestion?

internal/provider/common/adcdebugserver.go
Revolyssup: No. The underlying data store has adc configs translated by the ingress controller we're hitting. So obviously it will show those resources only.

config/samples/config.yaml
bzp2010: @AlinsRan @ronething Do you have any ideas? If not, let's use the `server` first and merge it.

ronething: i think it's okay to keep the `server` for now.

PR #2553, opened 2025-09-11
feat: add support for named servicePort in ApisixRoute backend

Revolyssup:
ApisixRoute schema supports int or string -https://github.com/apache/apisix-ingress-controller/blob/d3f8ce91dceba5a067379198d6872545a68042bb/samples/deploy/crd/v1/ApisixRoute.yaml#L210-L213 

But currently the code logic doesn't translate service port names for matching. The PR adds support for it.


AlinsRan (approved):

internal/adc/translator/apisixroute.go
```diff
@@ -364,14 +389,26 @@ func (t *Translator) translateApisixRouteStreamBackendResolveGranularity(tctx *p
...
 func (t *Translator) translateApisixRouteBackendResolveGranularityEndpoint(tctx *provider.TranslateContext, arNN types.NamespacedName, backend apiv2.ApisixRouteHTTPBackend) (adc.UpstreamNodes, error) {
+	serviceNN := types.NamespacedName{
+		Namespace: arNN.Namespace,
+		Name:      backend.ServiceName,
+	}
+	svc, ok := tctx.Services[serviceNN]
+	if !ok {
+		return nil, errors.Errorf("service not found, ApisixRoute: %s, Service: %s", arNN, serviceNN)
+	}
+	port, err := getPortFromService(svc, backend.ServicePort)
```
bzp2010, line 400: If I understand correctly, when we resolve ports using endpoints as the granularity, resolving the service port alone is insufficient.

We must further resolve the port used by the pod (i.e., service's targetPort) through the specified service port, and correctly handle cases where the targetPort is either a string or a number?
There is no guarantee that the service port and pod port will use the same value, so directly using the service port is inappropriate.

ping @AlinsRan @ronething

internal/adc/translator/apisixroute.go
```diff
@@ -328,6 +329,26 @@ func (t *Translator) buildService(ar *apiv2.ApisixRoute, rule apiv2.ApisixRouteH
...
 
+func getPortFromService(svc *v1.Service, backendSvcPort intstr.IntOrString) (int32, error) {
+	var port int32
+	if backendSvcPort.Type == intstr.Int {
+		port = int32(backendSvcPort.IntValue())
+	} else {
+		found := false
+		for _, servicePort := range svc.Spec.Ports {
+			if servicePort.Name == backendSvcPort.StrVal {
+				port = servicePort.Port
```
bzp2010, line 340: The `getPortFromService` directly uses `port` instead of `targetPort`.
When using the endpoint as the resolution granularity, this should not be the expected behavior?

bzp2010 (changes requested):
Block this merge temporarily until we reach a consensus on the concerns.

internal/adc/translator/apisixroute.go
Revolyssup: @bzp2010 Actually this has to be service port only. 
- In the case of endpoint granularity, we first generate the BackendRef object which is a service.
- Then translateBackendRef is called with the Service here https://github.com/apache/apisix-ingress-controller/blob/476783aab6090dbfb1051fc2ad021de5f490aada/internal/adc/translator/httproute.go#L378
- Then there are two cases further:
   1. Service was of type ExternalName. In this case, we dont need the endpointslices and simply return the UpstreamNode with serviceport and address https://github.com/apache/apisix-ingress-controller/blob/476783aab6090dbfb1051fc2ad021de5f490aada/internal/adc/translator/httproute.go#L402
   2. Service is not ExternalName and in this case we use the `portname` and `weight` from BackendRef to translate and filter out EndpointSlice for creating upstream nodes  https://github.com/apache/apisix-ingress-controller/blob/476783aab6090dbfb1051fc2ad021de5f490aada/internal/adc/translator/httproute.go#L425 

https://github.com/apache/apisix-ingress-controller/blob/476783aab6090dbfb1051fc2ad021de5f490aada/internal/adc/translator/httproute.go#L339

So in conclusion, the current logic is correct and here we only need the serviceport as endpoint will be later resolved in another function depending on service type

internal/adc/translator/apisixroute.go
Revolyssup: explained here - https://github.com/apache/apisix-ingress-controller/pull/2553/files#r2347360896

Revolyssup: The endpoints are still transparent to the user. We use BackendRef object to translate the `backends` and later depending on the granularity resolve endpoints in another function as described in the above comment.

bzp2010 (approved):

PR #2577, opened 2025-09-24
feat(apisixupstream): support discovery

AlinsRan:
### Type of change:

- [ ] Bugfix
- [x] New feature provided
- [ ] Improve performance
- [ ] Backport patches
- [ ] Documentation
- [ ] Refactor
- [ ] Chore
- [ ] CI/CD or Tests

The `ApisixUpstream.discovery` feature is already supported in version 1.x, but has not been implemented in version 2.x.

### Pre-submission checklist:

### DIFF
--- a/internal/adc/translator/apisixupstream.go
+++ b/internal/adc/translator/apisixupstream.go
@@ -42,10 +42,7 @@ func (t *Translator) translateApisixUpstream(tctx *provider.TranslateContext, au
 		translateApisixUpstreamRetriesAndTimeout,
 		translateApisixUpstreamPassHost,
 		translateUpstreamHealthCheck,
-<<<<<<< HEAD
 		translateUpstreamDiscovery,
-=======
->>>>>>> master
 	} {
 		if err = f(au, ups); err != nil {
 			return


internal/adc/translator/apisixroute.go
```diff
@@ -269,7 +269,7 @@ func (t *Translator) buildUpstream(tctx *provider.TranslateContext, service *adc
 	}
 
 	// no valid upstream
-	if len(upstreams) == 0 || len(upstreams[0].Nodes) == 0 {
+	if len(upstreams) == 0 {
```
ronething, line 272: Should we first check whether `Nodes` or `discovery type` exist before adding a single upstream to the upstreams array? If neither exists, should we not add it to the upstreams array?

AlinsRan: Why shouldn't it be?
Assuming that in a grayscale scenario, service a may be unavailable, but service b is available, I think this judgment is unreasonable.

ronething: That makes sense.

PR #2592, opened 2025-10-09
fix: generate unique SSL IDs to prevent certificate conflicts across different hosts

ronething:
### Type of change:

- [x] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches
- [ ] Documentation
- [ ] Refactor
- [ ] Chore
- [ ] CI/CD or Tests

Previously, SSL objects used certificate content to generate IDs (`id.GenID(string(cert))`).
When the same certificate was referenced by multiple resources (e.g., ApisixTls and Ingress) with different hostnames, they would generate the same SSL ID, causing conflicts where only one SSL object would be created and this may cause error. So we change the ssl id generate logic.

This ensures each TLS configuration creates a unique SSL object, even when using the same certificate.

And, because CI has been occurring errors, the tcproute test case was also fixed.

### Pre-submission checklist:

### DIFF
--- a/internal/adc/translator/gateway.go
+++ b/internal/adc/translator/gateway.go
@@ -120,7 +120,7 @@ func (t *Translator) translateSecret(tctx *provider.TranslateContext, listener g
 					}
 					sslObj.Snis = append(sslObj.Snis, hosts...)
 				}
-				sslObj.ID = id.GenID(fmt.Sprintf("Gateway_%s_%s_%s_%d", obj.Namespace, obj.Name, listener.Name, refIndex))
+				sslObj.ID = id.GenID(fmt.Sprintf("%s_%s_%d", adctypes.ComposeSSLName(internaltypes.KindGateway, obj.Namespace, obj.Name), listener.Name, refIndex))
 				t.Log.V(1).Info("generated ssl id", "ssl id", sslObj.ID, "secret", secretNN.String())
 				sslObj.Labels = label.GenLabel(obj)
 				sslObjs = append(sslObjs, sslObj)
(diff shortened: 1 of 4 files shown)


internal/adc/translator/gateway.go
```diff
@@ -122,8 +120,7 @@ func (t *Translator) translateSecret(tctx *provider.TranslateContext, listener g
 					}
 					sslObj.Snis = append(sslObj.Snis, hosts...)
 				}
-				// Note: use cert as id to avoid duplicate certificate across ssl objects
-				sslObj.ID = id.GenID(string(cert))
+				sslObj.ID = id.GenID(fmt.Sprintf("Gateway_%s_%s_%s_%d", obj.Namespace, obj.Name, listener.Name, refIndex))
```
AlinsRan: I think choosing 1 between `listener.name` and `refIndex` is enough

internal/adc/translator/ingress.go
```diff
@@ -64,7 +64,7 @@ func (t *Translator) translateIngressTLS(ingressTLS *networkingv1.IngressTLS, se
 		},
 		Snis: hosts,
 	}
-	ssl.ID = id.GenID(string(cert))
+	ssl.ID = id.GenID(fmt.Sprintf("Ingress_%s_%s_%d", namespace, name, tlsIndex))
```
AlinsRan: The name generation is universal, and I think it can be abstracted as the composeName function.

internal/adc/translator/gateway.go
AlinsRan: You're right, I noticed that index and lister are not at the same level.

internal/adc/translator/ingress.go
ronething: updated.

PR #2647, opened 2025-11-06
feat: support deployment of multiple data plane modes

AlinsRan:
### Type of change:

- [ ] Bugfix
- [x] New feature provided
- [ ] Improve performance
- [ ] Backport patches
- [ ] Documentation
- [ ] Refactor
- [ ] Chore
- [ ] CI/CD or Tests

By default, static configuration will prevail, and the deployment mode of the data plane can be dynamically specified.
`mode: apisix` or `mode: apisix-standalone`

```yaml
kind: GatewayProxy
metadata:
  name: apisix-proxy-config
spec:
  provider:
    type: ControlPlane
    controlPlane:
      mode: apisix # apisix-standalone
      service:
        name: hello
        port: 9180
      auth:
        type: AdminKey
        adminKey:
          value: "%s"
```

### Pre-submission checklist:

### DIFF
--- a/test/e2e/scaffold/apisix_deployer.go
+++ b/test/e2e/scaffold/apisix_deployer.go
@@ -324,55 +324,7 @@ func (s *APISIXDeployer) closeAdminTunnel() {
 }
 
 func (s *APISIXDeployer) CreateAdditionalGateway(namePrefix string) (string, *corev1.Service, error) {
-	// Create a new namespace for this additional gateway
-	additionalNS := fmt.Sprintf("%s-%d", namePrefix, time.Now().Unix())
-
-	k8s.CreateNamespace(s.t, s.kubectlOptions, additionalNS)
-
-	// Create new kubectl options for the new namespace
-	kubectlOpts := &k8s.KubectlOptions{
-		ConfigPath: s.runtimeOpts.Kubeconfig,
-		Namespace:  additionalNS,
-	}
-
-	s.Logf("additional gateway in namespace %s", additionalNS)
-
-	// Use the same admin key as the main gateway
-	adminKey := s.runtimeOpts.APISIXAdminAPIKey
-	s.Logf("additional gateway admin api key: %s", adminKey)
-
-	// Store gateway resources info
-	resources := &GatewayResources{
-		Namespace:   additionalNS,
-		AdminAPIKey: adminKey,
-	}
-
-	// Deploy dataplane for this additional gateway
-	opts := APISIXDeployOptions{
-		Namespace:        additionalNS,
-		AdminKey:         adminKey,
-		ServiceHTTPPort:  9080,
-		ServiceHTTPSPort: 9443,
-	}
-	svc := s.deployDataplane(&opts)
-
-	resources.DataplaneService = svc
-
-	// Create tunnels for the dataplane
-	tunnels, err := s.createDataplaneTunnels(svc, kubectlOpts, svc.Name)
-	if err != nil {
-		return "", nil, err
-	}
-
-	resources.Tunnels = tunnels
-
-	// Use namespace as identifier for APISIX deployments
-	identifier := additionalNS
-
-	// Store in the map
-	s.additionalGateways[identifier] = resources
-
-	return identifier, svc, nil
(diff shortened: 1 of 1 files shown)


api/adc/types.go
```diff
@@ -779,6 +779,7 @@ type Config struct {
 	ServerAddrs []string
 	Token       string
 	TlsVerify   bool
+	BakcnedType string
```
ronething: typo.

internal/adc/translator/gatewayproxy.go
```diff
@@ -44,47 +45,53 @@ func (t *Translator) TranslateGatewayProxyToConfig(tctx *provider.TranslateConte
...
-		config.ServerAddrs = endpoints
-		return &config, nil
+		cfg.ServerAddrs = endpoints
+		return &cfg, nil
 	}
 
-	if provider.ControlPlane.Service != nil {
+	if cp.Mode != "" {
+		resolveEndpoints = cp.Mode == string(config.ProviderTypeStandalone)
+	}
```
ronething, line 91: we can add some code comment.

test/e2e/scaffold/apisix_deployer.go
```diff
@@ -371,6 +375,77 @@ func (s *APISIXDeployer) CreateAdditionalGateway(namePrefix string) (string, *co
...
+		AdminKey:         adminKey,
+		ServiceHTTPPort:  9080,
+		ServiceHTTPSPort: 9443,
+	}
+	if opts.Namespace != "" {
+		o.Namespace = additionalNS
+	}
+	if opts.AdminKey != "" {
+		o.AdminKey = adminKey
+	}
```
ronething, line 366: Did you accidentally write it wrong or what?

api/adc/types.go
AlinsRan: fixed.

config/crd/bases/apisix.apache.org_gatewayproxies.yaml
```diff
@@ -150,6 +155,9 @@ spec:
                     type: object
                     x-kubernetes-validations:
                     - rule: has(self.endpoints) != has(self.service)
+                    - message: mode is immutable
+                      rule: oldSelf == null || (!has(self.mode) && !has(oldSelf.mode))
+                        || self.mode == oldSelf.mode
```
bzp2010, line 160: What is the aim here? What might occur if we do not prohibit this modification? 🤔

AlinsRan: 1. The primary conservative consideration is to prevent unforeseen consequences by introducing this restriction.

2. Since the ADC maintains in-memory state and different modes share the same cache_key, switching modes may affect the stored results.

bzp2010: 1. I understand that position.

However, please confirm: if I delete the GatewayProxy resource and recreate it with the same name using a different mode, no resources will be deleted in cascade, and the Ingress Controller will begin pushing configurations in the new mode, correct?

Once you have confirmed the issue, I shall approve and merge it. We will keep this restriction in place for the time being.

2. The cache will only exist on a single backend. It will not be shared across different backend instances. This is not a problem.

AlinsRan: 1. Yes, when deleting GatewayProxy, it will only delete the association relationship, not the routing resources.

https://github.com/apache/apisix-ingress-controller/blob/6f0e78cfe8a702ef66fa8c448ee5381f8422e430/internal/controller/gatewayproxy_controller.go#L84-L91

2. So the cache of ADC is distinguished by cache_key+type?

bzp2010: 2. 
Essentially, yes. Each backend establishes its own cache based on its requirements and maintains it internally without sharing.
Currently, only the apisix-standalone backend incorporates caching; the other backend types do not utilise caching.

The type will be determined by the `backend` value you input during the sync operation and does not require explicit specification in the `cacheKey`. Naturally, given that the `cacheKey` is a transparent string, you may also specify it explicitly if desired.

PR #2693, opened 2026-01-06
chore: log when kubernetes cluster version does not meet minimum requirement

ronething:
### Type of change:

- [ ] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches
- [ ] Documentation
- [ ] Refactor
- [x] Chore
- [ ] CI/CD or Tests

1. Optimize the ingress documentation and clearly inform the reasons for relying on the 1.26+ cluster version

2. When ingress starts and detects that the cluster version does not meet the requirements, it prints a warning log to alert the user

### Pre-submission checklist:

### DIFF
--- a/internal/manager/run.go
+++ b/internal/manager/run.go
@@ -55,6 +55,11 @@ var (
 	scheme = runtime.NewScheme()
 )
 
+var (
+	// set value during compilation
+	_minK8sVersion string
+)
+
 func init() {
 	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
 
@@ -261,8 +266,7 @@ func Run(ctx context.Context, logger logr.Logger) error {
 }
 
 func checkK8sVersion(mgr ctrl.Manager, logger logr.Logger) {
-	const minVersion = "1.26.0"
-	minV, err := version.ParseSemantic(minVersion)
+	minV, err := version.ParseSemantic(_minK8sVersion)
 	if err != nil {
 		logger.Info("failed to parse minimum version", "error", err)
 		return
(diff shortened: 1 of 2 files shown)


internal/manager/run.go
```diff
@@ -254,3 +259,37 @@ func Run(ctx context.Context, logger logr.Logger) error {
...
+	}
+
+	currentVersion, err := version.ParseSemantic(serverVersion.GitVersion)
+	if err != nil {
+		logger.Info("failed to parse server version", "error", err)
+		return
+	}
+
+	if !currentVersion.AtLeast(minV) {
+		logger.Info("WARNING: Kubernetes cluster version does not meet minimum requirement",
```
nic-6443, line 294: Should the log level also be warn?

ronething: github.com/go-logr/logr does not have a direct warn log level.

internal/manager/run.go
```diff
@@ -254,3 +259,37 @@ func Run(ctx context.Context, logger logr.Logger) error {
 	setupLog.Info("starting controller manager")
 	return mgr.Start(ctrl.SetupSignalHandler())
 }
+
+func checkK8sVersion(mgr ctrl.Manager, logger logr.Logger) {
+	const minVersion = "1.26.0"
```
AlinsRan: Maybe we can pass parameters during code compilation. What do you think? Similar to version.

nic-6443 (approved):

ronething:
Test cases that cause errors are not within the scope of local modifications.

internal/manager/run.go
ronething: updated.

PR #2743, opened 2026-04-10
fix: remove redundant route.Hosts to prevent false diffs in ADC sync

nic-6443:
## Problem

The translator sets `hosts` on both Route and Service for ApisixRoute resources. Since APISIX routes inherit hosts from their parent service, the route-level hosts is redundant.

For backends that don't support route-level hosts, this causes a false diff every sync cycle: the local state has `route.hosts` but the remote state never does, triggering unnecessary PUT requests and audit log bloat.

## Root Cause

In `buildRoute()` (`internal/adc/translator/apisixroute.go`), `route.Hosts` was set to `rule.Match.Hosts`. However, `buildService()` already sets `service.Hosts` to the same value. Since routes inherit hosts from their parent service, the route-level hosts is redundant and can cause false diffs during config sync.

## Fix

Remove `route.Hosts = rule.Match.Hosts` from `buildRoute()`. `service.Hosts` remains as the canonical location for host matching.

## CI Fix

Also fixes ADC binary extraction in CI workflows: the `ghcr.io/api7/adc:dev` Docker image now bundles `main.cjs` instead of `main.js` at the container root. Updated `apisix-e2e-test.yml` and `benchmark-test.yml`.

## Testing

Added unit tests:
- `TestBuildRoute_HostsNotSet`: verifies route.Hosts is not set after buildRoute
- `TestBuildService_HostsSet`: verifies service.Hosts is correctly set

### DIFF
--- a/.github/workflows/apisix-e2e-test.yml
+++ b/.github/workflows/apisix-e2e-test.yml
@@ -95,7 +95,7 @@ jobs:
         if: ${{ env.ADC_VERSION == 'dev' }}
         run: |
           docker create --name adc-temp ghcr.io/api7/adc:dev
-          docker cp adc-temp:main.js adc.js
+          docker cp adc-temp:main.cjs adc.js
           docker rm adc-temp
           node $(pwd)/adc.js -v
           echo "ADC_BIN=node $(pwd)/adc.js" >> $GITHUB_ENV
@@ -165,7 +165,7 @@ jobs:
         if: ${{ env.ADC_VERSION == 'dev' }}
         run: |
           docker create --name adc-temp ghcr.io/api7/adc:dev
-          docker cp adc-temp:main.js adc.js
+          docker cp adc-temp:main.cjs adc.js
           docker rm adc-temp
           node $(pwd)/adc.js -v
           echo "ADC_BIN=node $(pwd)/adc.js" >> $GITHUB_ENV
--- a/.github/workflows/benchmark-test.yml
+++ b/.github/workflows/benchmark-test.yml
@@ -98,7 +98,7 @@ jobs:
         if: ${{ env.ADC_VERSION == 'dev' }}
         run: |
           docker create --name adc-temp ghcr.io/api7/adc:dev
-          docker cp adc-temp:main.js adc.js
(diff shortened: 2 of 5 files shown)


nic-6443:
**CI Status**: lint ✅, unit tests ✅, misspell ✅, CRD docs ✅, license check ✅

The e2e test failures are a **pre-existing infrastructure issue** — the ADC dev Docker image (`ghcr.io/api7/adc:dev`) no longer has `main.js` at the container root, causing the "Extract adc binary" step to fail with `Could not find the file main.js in container adc-temp`. This is unrelated to the changes in this PR.

nic-6443:
Pushed CI fix: updated `main.js` → `main.cjs` in `apisix-e2e-test.yml` and `benchmark-test.yml` to match the updated `ghcr.io/api7/adc:dev` Docker image layout. The e2e tests should now pass the ADC binary extraction step.

nic-6443:
**CI Status update**: All APISIX E2E Tests ✅ (7/7 jobs passed), all conformance tests ✅, lint ✅, unit tests ✅, CodeQL ✅, license ✅.

The only failure is `e2e-test (apisix-standalone)` in the **Benchmark Test** workflow — APISIX returns 500 Internal Server Error on `ListServices` during test scaffold setup. This is an infrastructure issue (APISIX standalone not properly initialized), unrelated to the route.Hosts removal in this PR.

nic-6443:
Pushed e2e test fix: replaced 2 direct HTTP assertions with `Eventually` retry in `test/e2e/crds/v2/route.go`.

**Root cause**: In APISIX standalone mode, config application via `/apisix/admin/configs` is async (202 response). The Basic test and WebSocket test asserted HTTP status immediately after config push without retry, causing intermittent 404s when APISIX hadn't applied the new config yet.

**Fix**: Use `Eventually` with 20s timeout + 1s polling (same pattern as other assertions in the test suite).

The previous Benchmark test failure (APISIX 500 on ListServices) was an infrastructure issue — APISIX standalone failed to initialize. This should resolve on re-run.

AlinsRan (approved):

shreemaan-abhishek (approved):

PR #2750, opened 2026-04-22
docs: remove redundant namespace from HTTPRoute, Ingress, and ApisixRoute examples in configure-routes.md

fengqi-dev:
### Type of change:

- [ ] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches
- [x] Documentation
- [ ] Refactor
- [ ] Chore
- [ ] CI/CD or Tests

This pull request makes minor updates to the documentation in `docs/en/latest/getting-started/configure-routes.md` by removing the `namespace: ingress-apisix` field from several Kubernetes resource examples. 
https://github.com/apache/apisix-ingress-controller/blob/878e5015bda978cb2f4231511c835420d8769a2f/examples/httpbin/deployment.yaml#L60-L63

### Pre-submission checklist:

### DIFF
--- a/docs/en/latest/getting-started/configure-routes.md
+++ b/docs/en/latest/getting-started/configure-routes.md
@@ -134,7 +134,6 @@ values={[
 apiVersion: gateway.networking.k8s.io/v1
 kind: HTTPRoute
 metadata:
-  namespace: ingress-apisix
   name: getting-started-ip
 spec:
   parentRefs:
@@ -157,7 +156,6 @@ spec:
 apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
-  namespace: ingress-apisix
   name: getting-started-ip
 spec:
   ingressClassName: apisix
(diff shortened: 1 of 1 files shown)


Baoyuantop (approved):

kayx23:
I don't think this change is safe as-is. Removing `metadata.namespace` here changes behavior, not just style.

For the `HTTPRoute` example, the doc defines the `Gateway` in `ingress-apisix` earlier in the same guide. If `httpbin-route.yaml` is applied without an explicit namespace, it will be created in the user's current namespace, so `parentRefs: - name: apisix` may no longer refer to the `Gateway` in `ingress-apisix`. That can break the Gateway API walkthrough.

The same issue applies to the Ingress and `ApisixRoute` examples: without `metadata.namespace`, the examples become dependent on the user's current kubectl context. The guide currently does not set a default namespace and does not use `-n ingress-apisix` when applying the route manifest.

I think we should either:
1. keep the explicit `namespace: ingress-apisix` in these examples, or
2. update the whole walkthrough to set/apply a namespace consistently (`kubectl config set-context --current --namespace=ingress-apisix` or `kubectl apply -n ingress-apisix ...`).

Right now this PR makes the docs less explicit and potentially incorrect for the Gateway API case.

PR #2758, opened 2026-05-06
feat: add ADC-backed admission validation for APISIX CRDs

AlinsRan:
## Summary

This PR adds live ADC-backed admission validation to the existing webhook validators for `ApisixRoute`, `ApisixConsumer`, `ApisixTls`, and `Consumer` resources.

Previously the webhooks only emitted warnings for missing references (services, secrets). With this change, resources are also structurally validated against a live APISIX instance before being admitted.

### Core infrastructure

- **`internal/types/error.go`**: Add `ADCValidationErrors`, `ADCValidationError`, `ADCValidationServerAddrError`, and `ADCValidationDetail` types to carry structured validation error details.

- **`internal/adc/client/executor.go`**: 
  - Add `Validate()` to the `ADCExecutor` interface and `HTTPADCExecutor`
  - Implement `runHTTPValidate` / `runHTTPValidateForSingleServer` that POST to the `/configs/validate` endpoint
  - Refactor `buildHTTPRequest` to accept an HTTP method and path (supporting both `/sync` and `/configs/validate`)
  - Set TLS minimum version to 1.2
  - Redact full request body from logs (log length only)

- **`internal/adc/client/client.go`**: Add `Client.Validate()` which calls the executor's validate path and aggregates `ADCValidationErrors`.

### Webhook validation helpers

- **`internal/controller/webhook_validation.go`** *(new)*: Lightweight `Prepare*ForValidation` helpers for each CRD type that build a `TranslateContext` without running the full reconciler loop. Used by the admission webhook to resolve references before translating.

### Admission validator

- **`internal/webhook/v1/adc_validation.go`** *(new)*: `adcAdmissionValidator` that:
  - Resolves the IngressClass / GatewayProxy for the resource
  - Translates the resource into an ADC payload
  - Posts the payload to APISIX via `client.Validate()`
  - **Fails open** on infrastructure / transport errors (only `ADCValidationErrors` cause denial)
  - Populates `global_rules` and `plugin_metadata` in the validate payload so plugin references can be resolved

### Webhook wiring

- **`apisixroute_webhook.go`**, **`apisixconsumer_webhook.go`**, **`apisixtls_webhook.go`**, **`consumer_webhook.go`**: Wire `adcAdmissionValidator`; ADC init errors are logged and ignored (fail-open).

- **`apisixtls_webhook.go`**: Skip ADC validation when secrets are missing to preserve the existing warn-only behaviour for that case.

- **`consumer_webhook.go`**: Validate duplicate `key-auth` credential keys scoped to the same `GatewayRef` using a field-index query (O(1) instead of O(N) full list). Malformed inline JSON credentials are logged and skipped rather than causing a hard denial.

## Behaviour notes

| Scenario | Before | After |
|---|---|---|
| Valid resource, ADC reachable | Admit with warnings | Admit (pass validation) |
| Invalid resource, ADC reachable | Admit with warnings | Deny with structured errors |
| ADC unreachable / init error | Admit with warnings | Admit with warnings (fail-open) |
| ApisixTls / ApisixConsumer with missing secrets | Admit with warnings | Admit with warnings (unchanged) |
| Consumer with duplicate key-auth key | Admit | Deny |
| Consumer with malformed inline credential JSON | Admit | Admit (log + skip duplicate check) |

### Deny: invalid plugin configuration on ApisixRoute

Applying an `ApisixRoute` whose `response-rewrite` plugin receives `status_code` as a string instead of an integer is rejected immediately:

```yaml
kind: ApisixRoute
metadata:
  name: my-route
  namespace: default
spec:
  ingressClassName: apisix
  http:
  - name: rule1
    match:
      hosts: [example.com]
      paths: [/api]
    backends:
    - serviceName: my-svc
      servicePort: 80
    plugins:
    - name: response-rewrite
      enable: true
      config:
        status_code: "500"   # wrong type: must be integer
```

```
$ kubectl apply -f route.yaml
Error from server: admission webhook "vapisixroute.kb.io" denied the request:
  ADC validation errors: [ADC validation error for ApisixRoute/default/my-route:
    [ServerAddr: 127.0.0.1:9180, Err: route rejected
      (type=route, name=my-route, property 'status_code' validation failed: wrong type: expected integer, but got string)]]
```

### Deny: invalid jwt-auth algorithm on ApisixConsumer

```yaml
kind: ApisixConsumer
metadata:
  name: my-consumer
  namespace: default
spec:
  ingressClassName: apisix
  authParameter:
    jwtAuth:
      value:
        key: my-consumer-key
        algorithm: INVALID_ALGO   # must be one of RS256, RS512, ES256, ES512, HS256, HS512
        private_key: |
          -----BEGIN RSA PRIVATE KEY-----
          ...
          -----END RSA PRIVATE KEY-----
```

```
$ kubectl apply -f consumer.yaml
Error from server: admission webhook "vapisixconsumer.kb.io" denied the request:
  ADC validation errors: [ADC validation error for ApisixConsumer/default/my-consumer:
    [ServerAddr: 127.0.0.1:9180, Err: consumer rejected
      (type=consumer, name=my-consumer, plugin jwt-auth: property 'algorithm' validation failed: value must be one of RS256, RS512, ES256, ES512, HS256, HS512)]]
```

### Warn + Admit: missing backend Service on ApisixRoute

When a referenced Service does not yet exist, the resource is **admitted** with a warning (eventual consistency):

```yaml
kind: ApisixRoute
metadata:
  name: my-route
  namespace: default
spec:
  ingressClassName: apisix
  http:
  - name: rule1
    match:
      hosts: [example.com]
      paths: [/api]
    backends:
    - serviceName: not-yet-created-svc   # service does not exist yet
      servicePort: 80
```

```
$ kubectl apply -f route.yaml
Warning: Referenced Service 'default/not-yet-created-svc' not found
apisixroute.apisix.apache.org/my-route created
```

### Deny: SNI conflict on ApisixTls

When two `ApisixTls` resources claim the same SNI, the second one is denied:

```yaml
# already exists
kind: ApisixTls
metadata:
  name: tls-a
  namespace: default
spec:
  ingressClassName: apisix
  hosts: [example.com]
  secret:
    name: tls-secret-a
    namespace: default
---
# new resource — same host
kind: ApisixTls
metadata:
  name: tls-b
  namespace: default
spec:
  ingressClassName: apisix
  hosts: [example.com]   # conflicts with tls-a
  secret:
    name: tls-secret-b
    namespace: default
```

```
$ kubectl apply -f tls-b.yaml
Error from server: admission webhook "vapisixtls.kb.io" denied the request:
  SNI "example.com" is already used by ApisixTls default/tls-a
```


nic-6443 (approved):

internal/controller/webhook_validation.go
```diff
@@ -0,0 +1,113 @@
...
+		Client: c,
+		Log:    log,
+	}
+	if err := reconciler.processSpec(ctx, tctx, consumer); err != nil {
+		return nil, err
+	}
+	return tctx, nil
+}
+
+func PrepareConsumerForValidation(ctx context.Context, c client.Client, log logr.Logger, consumer *v1alpha1.Consumer) (*provider.TranslateContext, error) {
```
shreemaan-abhishek, line 74: why are there two similar functions `PrepareApisixConsumerForValidation` and `PrepareConsumerForValidation`? and this file does not contain functions for all apisix resources, why?

internal/webhook/v1/apisixroute_webhook.go
```diff
@@ -41,19 +41,24 @@ func SetupApisixRouteWebhookWithManager(mgr ctrl.Manager) error {
 		Complete()
 }
 
-// +kubebuilder:webhook:path=/validate-apisix-apache-org-v2-apisixroute,mutating=false,failurePolicy=fail,sideEffects=None,groups=apisix.apache.org,resources=apisixroutes,verbs=create;update,versions=v2,name=vapisixroute-v2.kb.io,admissionReviewVersions=v1,failurePolicy=Ignore
+// +kubebuilder:webhook:path=/validate-apisix-apache-org-v2-apisixroute,mutating=false,failurePolicy=Ignore,sideEffects=None,groups=apisix.apache.org,resources=apisixroutes,verbs=create;update,versions=v2,name=vapisixroute-v2.kb.io,admissionReviewVersions=v1
```
shreemaan-abhishek, line 44: now why do we just ignore in failure?

internal/controller/webhook_validation.go
AlinsRan: There are two types of resource `kind: Consumer` and `kind: ApisixConsumer`

shreemaan-abhishek (approved):

internal/webhook/v1/apisixroute_webhook.go
AlinsRan: Ignoring it is intentional. The current installation is designed to ignore this case; the only issue here is that the annotation is incorrect.

If we do not ignore it, then when the controller is unavailable, it may prevent resources from being admitted or taking effect at all. For example, during startup,
before the controller is ready, it could intercept standard resources such as Ingress and IngressClass, even though those resources may be managed by other controllers. That would not be the expected behavior.

PR #2761, opened 2026-05-09
feat: support plugins field in ApisixConsumer

AlinsRan:
## Summary

Add a generic `plugins` field to `ApisixConsumerSpec` so that consumer-scoped plugins (e.g. `limit-count`, `limit-req`) can be attached to an `ApisixConsumer` resource directly, without being limited to the auth plugins exposed through `authParameter`.

## Changes

- **`ApisixConsumerSpec`** gains a `Plugins []ApisixRoutePlugin` field, following the same pattern as `ApisixRoute` and `ApisixGlobalRule`. Enabled plugins are merged after the auth plugin derived from `authParameter`; an enabled plugin with the same name takes precedence.
- **`authParameter` is now optional** (`omitempty`). Both `authParameter` and `plugins` are optional at the API level; the controller handles the case where neither is set.
- **Translator** updated to process the new `Plugins` slice via the existing `buildPluginConfig` helper.
- **Controller and indexer** extended to load and index Secrets referenced by `spec.plugins[].secretRef`, consistent with `ApisixRoute` and `ApisixGlobalRule` behavior.
- **deepcopy** updated for the new `Plugins` slice and now-optional `AuthParameter` pointer (auto-generated by `make generate`).
- **E2e tests** added for two scenarios:
  - `authParameter` + extra `plugins` (e.g. `limit-count` rate limiting at consumer level)
  - `plugins` only — no `authParameter` (e.g. `key-auth` configured directly via plugins)

## Example

```yaml
kind: ApisixConsumer
metadata:
  name: my-consumer
spec:
  ingressClassName: apisix
  authParameter:
    keyAuth:
      value:
        key: my-secret-key
  plugins:
  - name: limit-count
    enable: true
    config:
      count: 100
      time_window: 60
      rejected_code: 429
      policy: local
```

Or, using `plugins` only (without `authParameter`):

```yaml
kind: ApisixConsumer
metadata:
  name: key-auth-consumer
spec:
  ingressClassName: apisix
  plugins:
  - name: key-auth
    enable: true
    config:
      key: my-secret-key
```


nic-6443 (approved):

internal/adc/translator/apisixconsumer.go
```diff
@@ -55,42 +55,54 @@ const (
...
-		if err != nil {
-			return nil, fmt.Errorf("invalid key auth config: %s", err)
+	if ap := ac.Spec.AuthParameter; ap != nil {
+		if ap.KeyAuth != nil {
+			cfg, err := t.translateConsumerKeyAuthPlugin(tctx, ac.Namespace, ap.KeyAuth)
+			if err != nil {
+				return nil, fmt.Errorf("invalid key auth config: %s", err)
+			}
+			plugins["key-auth"] = cfg
+		} else if ap.BasicAuth != nil {
```
shreemaan-abhishek, line 65: shouldn't we support multi-auth too?

shreemaan-abhishek (approved):

internal/adc/translator/apisixconsumer.go
AlinsRan: The purpose of this PR is to support plugins.
`multi-auth` can still be configured through plugins.

PR #2768, opened 2026-05-20
fix: update observedGeneration on Gateway-API route status conditions

### LINKED ISSUE #2767: bug: ObservedGeneration not stamped on Gateway-API route status conditions
### Current Behavior

`apisix-ingress-controller` writes status conditions on `HTTPRoute`, `UDPRoute`, `TCPRoute`, and `GRPCRoute` parent rows with `observedGeneration: 0` instead of the live object's `metadata.generation`. The `metav1.Condition` convention is that `observedGeneration` reflects the spec generation the controller processed; consumers compare it to `metadata.generation` to tell whether the status reflects the current spec.

This behavior is not exhibited in the old CRDs like ApisixRoute, they are in line with the standard behavior.

### Expected Behavior

APISIX IC should set the route's `observedGeneration` to the generation it observes. Notably, this is already the case with the old CRDs (such as ApisixRoute)

- [`k8s.io/apimachinery/pkg/apis/meta/v1.Condition`](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#Condition)

  > `ObservedGeneration` represents the `.metadata.generation` that the condition was set based upon. For instance, if `.metadata.generation` is currently 12, but the `.status.conditions[x].observedGeneration` is 9, the condition is out of date with respect to the current state of the instance.

- [Kubernetes API conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties) — controllers must stamp `observedGeneration` from `metadata.generation` when reconciling the spec.

### Error Logs

_No response_

### Steps to Reproduce

1. Deploy a Gateway and an HTTPRoute (parent attached to the Gateway).
2. Trigger an ADC sync failure, then restore — for instance, point the APISIX LB at an invalid `ExternalName`, then set it to the right one again.
3. Inspect the route's `observedGeneration` on the parent's `Accepted` condition:
   ```yaml
   metadata:
     generation: 1
   status:
     parents:
     - conditions:
       - type: Accepted
         status: "True"
         reason: Accepted
         observedGeneration: 0   # ← should be 1
   ```

### Environment

- `apisix-ingress-controller` master or 2.0.1
- Kubernetes 1.30

hyarkoni-DN:
### Type of change:

- [x] Bugfix
- [ ] New feature provided
- [ ] Improve performance
- [ ] Backport patches
- [ ] Documentation
- [ ] Refactor
- [ ] Chore
- [ ] CI/CD or Tests

Fixes #2767.

Stamps `condition.ObservedGeneration = cp.GetGeneration()` inside the four Gateway-API mutator closures (`HTTPRoute`, `UDPRoute`, `TCPRoute`, `GRPCRoute`) before `MergeCondition`, mirroring what `SetApisixCRDConditionWithGeneration` already does for the legacy CRDs.

**Before** (HTTPRoute after a sync failure→recovery cycle):

```yaml
metadata:
  generation: 1
status:
  parents:
  - conditions:
    - type: Accepted
      status: "True"
      observedGeneration: 0   # mismatched
```

**After:**

```yaml
metadata:
  generation: 1
status:
  parents:
  - conditions:
    - type: Accepted
      status: "True"
      observedGeneration: 1   # matches metadata.generation
```

### Pre-submission checklist:

- [ ] Have you modified the corresponding document? (N/A)

### DIFF
--- a/internal/provider/apisix/status.go
+++ b/internal/provider/apisix/status.go
@@ -123,6 +123,7 @@ func (d *apisixProvider) updateStatus(nnk types.NamespacedNameKind, condition me
 			Resource:       &gatewayv1.HTTPRoute{},
 			Mutator: status.MutatorFunc(func(obj client.Object) client.Object {
 				cp := obj.(*gatewayv1.HTTPRoute).DeepCopy()
+				condition.ObservedGeneration = cp.GetGeneration()
 				gatewayNs := cp.GetNamespace()
 				for i, ref := range cp.Status.Parents {
 					ns := gatewayNs
@@ -158,6 +159,7 @@ func (d *apisixProvider) updateStatus(nnk types.NamespacedNameKind, condition me
 			Resource:       &gatewayv1alpha2.UDPRoute{},
 			Mutator: status.MutatorFunc(func(obj client.Object) client.Object {
 				cp := obj.(*gatewayv1alpha2.UDPRoute).DeepCopy()
+				condition.ObservedGeneration = cp.GetGeneration()
 				gatewayNs := cp.GetNamespace()
 				for i, ref := range cp.Status.Parents {
 					ns := gatewayNs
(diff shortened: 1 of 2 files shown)


AlinsRan (approved):
LGTM

nic-6443 (approved):

shreemaan-abhishek (approved):

PR #2794, opened 2026-06-18
feat: Add excludeResourceType config to prevent ADC from deleting API managed resources

luarx:
### Type of change:

- [ ] Bugfix
- [x] New feature provided
- [ ] Improve performance
- [ ] Backport patches
- [ ] Documentation
- [ ] Refactor
- [ ] Chore
- [ ] CI/CD or Tests

Introduces a new `provider.exclude_resource_type` config field (e.g. `[Consumer, ConsumerGroup]`) that is threaded through ProviderConfig → provider.Options → adcclient.New → Client.Sync(), so periodic ADC reconciliation sweeps skip resources of those types that were created via API rather than CRD — preventing the controller from deleting them.

Implements this issue: https://github.com/apache/apisix-ingress-controller/issues/2795

 ▎--> And will enable this PR to be merged if accepted https://github.com/apache/apisix-helm-chart/pull/983 

### Pre-submission checklist:

▎ Backward compatible: Yes — exclude_resource_type defaults to []/unset. Options.ApplyToList only overrides ExcludeResourceType when len(o.ExcludeResourceType) > 0, so existing configs behave exactly as before.

### DIFF
--- a/config/samples/config.yaml
+++ b/config/samples/config.yaml
@@ -49,6 +49,11 @@ provider:
                                         # If you want to enable the sync, set it to a positive value.
   init_sync_delay: 20m                  # The initial delay before the first sync, only used when the controller is started.
                                         # The default value is 20 minutes.
+  exclude_resource_type: []             # Resource types to exclude from full reconciliation sweeps.
+                                        # Resources of these types that exist in APISIX but have no corresponding
+                                        # CRD will not be deleted during periodic syncs.
+                                        # Supported values mirror the ADC resource types, e.g.: Consumer, ConsumerGroup.
+                                        # The default value is [] (empty — all resource types are reconciled).
 
 webhook:
   enable: false                         # Whether to enable the webhook server.
--- a/docs/en/latest/reference/configuration-file.md
+++ b/docs/en/latest/reference/configuration-file.md
@@ -78,4 +78,10 @@ provider:
                                         # If you want to enable the sync, set it to a positive value.
   init_sync_delay: 20m                  # The initial delay before the first sync, only used when the controller is started.
                                         # The default value is 20 minutes.
+
+  exclude_resource_type: []             # Resource types to exclude from full reconciliation sweeps.
+                                        # Resources of these types that exist in APISIX but have no corresponding
+                                        # CRD will not be deleted during periodic syncs.
+                                        # Supported values mirror the ADC resource types, e.g.: Consumer, ConsumerGroup.
+                                        # The default value is [] (empty — all resource types are reconciled).
 ```
--- a/internal/adc/client/client.go
+++ b/internal/adc/client/client.go
@@ -129,12 +129,13 @@ func isConfVersionRejection(err error) bool {
 const confVersionField = "conf_version"
 
 type Task struct {
-	Key           types.NamespacedNameKind
-	Name          string
-	Labels        map[string]string
-	Configs       map[types.NamespacedNameKind]adctypes.Config
-	ResourceTypes []string
-	Resources     *adctypes.Resources
+	Key                 types.NamespacedNameKind
+	Name                string
+	Labels              map[string]string
+	Configs             map[types.NamespacedNameKind]adctypes.Config
+	ResourceTypes       []string
+	ExcludeResourceType []string
+	Resources           *adctypes.Resources
 }
 
 // MarshalLog implements logr.Marshaler so logging a Task never dumps the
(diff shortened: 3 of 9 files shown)


luarx:
Friendly ping on this 👋 Has anyone had a chance to review it yet? 🙏

I'd really appreciate any feedback @nic-6443 @shreemaan-abhishek @AlinsRan, if one of you has some time, I'd love to get this moving as it is blocking us and maybe you know any other solution 😇

Btw, I am the same one that opened this unrelated [merged PR ](https://github.com/apache/apisix/pull/13616) too :)

luarx:
Friendly ping on this 👋 Has anyone had a chance to review it yet? 🙏

AlinsRan:
As written, `exclude_resource_type` doesn't skip garbage collection — it turns off syncing for that type entirely. I don't think a per-type opt-out is the right place to solve this.

**1. Excluding a type disables it, not just its GC.**

With `[consumer]` set, `Consumer`/`ApisixConsumer` CRDs would still reconcile and report a healthy status, but would never reach APISIX. Deleting a Gateway would strand the CRD-managed consumers there permanently.

Two things combine to cause that:

- `Sync()` isn't a periodic sweep sitting beside a per-CRD write path — it *is* the write path. `apisixProvider.Update()` calls `client.UpdateConfig()`, which only writes the local Store, then leans on `defer d.syncNotify()` to wake the loop in `Start()`, which calls `client.Sync()`. `Client.Update()`, the one that actually pushes, has no callers.
- ADC's `filterResourceType` strips the excluded type from both local and remote before diffing (`apps/cli/src/server/sync.ts`), so the type drops out of the diff altogether rather than just being spared deletion.

**2. The root cause looks like the missing label selector on the full sync.**

The `Task` built in `Client.Sync()` leaves `Labels` nil, so ADC diffs against the entire remote config and ignores the `manager-by: apisix-ingress-controller` label the translators already put on services, SSLs and consumers.

Scoping the sweep to that selector would protect externally created resources of any type — no new config knob, and no turning off reconciliation for the types you exclude. Global rules and plugin metadata have no labels to select on, but ADC's `filterConfiguration` already skips both, so they'd keep diffing as they do today.

It isn't free: anything created by a controller old enough to predate the label would suddenly look unmanaged. That needs verifying before anyone commits to it. But it goes at the real problem, and I'd rather we try that first.

**3. The scenario doesn't arise under standalone.**

`apisix-standalone` has no per-resource Admin API, just GET/PUT on `/apisix/admin/configs`, so nothing external can create a consumer out of band there — while the side effect in point 1 still applies.

---

One small thing for whenever this settles: ADC resource types are lowercase snake_case (`route`, `service`, `ssl`, `global_rule`, `consumer`, `consumer_group`, ...), so `Consumer` and `ConsumerGroup` in the sample config and docs would be rejected by the server's schema validation.

AlinsRan:
**1. The write path this PR assumes doesn't exist.**

The design assumes two independent paths: (a) a CRD changes and the controller pushes that resource, and (b) a periodic sweep that deletes whatever has no CRD behind it. `exclude_resource_type` is meant to touch only (b).

Path (a) isn't there. `apisixProvider.Update()` calls `client.UpdateConfig()`, which writes the local Store and nothing else — it sends no request to APISIX. It then relies on `defer d.syncNotify()` to wake the loop in `Start()`, which calls `client.Sync()`. `Client.Update()`, the only method that actually pushes, has no callers anywhere in the repo. Every CRD change is delivered by (b); there is no separate sweep to open a hole in.

And ADC implements the exclusion by dropping the type from both the local and the remote configuration before diffing (`filterResourceType`, called twice in `apps/cli/src/server/sync.ts`). With neither side carrying consumers, the diff has no consumer events at all — nothing gets deleted, but nothing gets created or updated either.

Put together, here is what someone running `exclude_resource_type: [consumer]` would see: they create a Consumer CRD, it reconciles, the status is healthy, the controller logs no error — and the consumer never reaches APISIX. Deleting a Gateway leaves the CRD-managed consumers stranded there with nothing left to clean them up. That's a worse failure than the one being fixed, and a silent one.

**2. The deletions come from the sweep not knowing what it owns.**

`Client.Sync()` builds its `Task` with `Labels` nil, which tells ADC to diff the local config against *everything* in APISIX and delete whatever the local side doesn't have. That's why externally created consumers disappear — not because the controller lacks a per-type opt-out.

The translators already stamp `manager-by: apisix-ingress-controller` on services, SSLs and consumers; the sweep just doesn't use it. Passing it as the label selector changes the question from "what isn't in my config?" to "what did I create that isn't in my config?", which is the one that should have been asked. That protects externally created resources of every type, needs no new configuration surface, and doesn't switch off reconciliation for anything. Global rules and plugin metadata have no labels to select on, but ADC's `filterConfiguration` already skips both, so they keep diffing exactly as they do today.

The cost is that resources created by a controller predating the label would suddenly look unowned — no longer updated, no longer collected. That needs checking against real deployments before committing to it.

**3. Under standalone the problem being solved doesn't exist.**

`apisix-standalone` exposes only GET/PUT on `/apisix/admin/configs`, with no per-resource Admin API, so nothing outside the controller can create a consumer there in the first place. The scenario this PR targets can't arise in that mode — while the silent failure from point 1 still can, and `validateProvider()` doesn't distinguish the two providers.

AlinsRan:
I've left a design review on the issue this implements: https://github.com/apache/apisix-ingress-controller/issues/2795#issuecomment-5502297588

PR #2801, opened 2026-07-07
fix: don't require certificateRefs for Passthrough-mode Gateway listeners

ehncraft:
### Type of change:

- [x] Bugfix

`translateSecret()` rejects any TLS listener with `nil` `CertificateRefs` before it even checks `listener.TLS.Mode`, so any Gateway with a `Passthrough`-mode listener fails translation with:

```
no certificateRefs found in listener <name>
```

and the Gateway's `Accepted` condition is permanently `False` — even though `CertificateRefs` must be empty in `Passthrough` mode per the Gateway API spec (TLS is never terminated at the Gateway, so there's no certificate for it to present). The `switch` statement already has a case correctly documenting this:

```go
// Only supported on TLSRoute. The certificateRefs field is ignored in this mode.
case gatewayv1.TLSModePassthrough:
	return sslObjs, nil
```

but the guard above the switch makes that case unreachable.

This also isn't purely cosmetic: `TranslateGateway` returns early on this error, so it never reaches the `GatewayProxy` global-rules/plugin-metadata aggregation at the bottom of the function — meaning any cluster-wide plugins/plugin-metadata configured via a `GatewayProxy` attached to a Gateway with a Passthrough listener silently never get applied.

This PR moves the nil check into the `Terminate` case, the only mode that actually needs a certificate.

### Pre-submission checklist:

- [x] Is this PR backward compatible? Yes — it only relaxes a previously-unconditional error for a spec-valid input (`Passthrough` listener with no `certificateRefs`). `Terminate`-mode behavior is unchanged.

### DIFF
--- a/internal/adc/translator/gateway.go
+++ b/internal/adc/translator/gateway.go
@@ -72,12 +72,12 @@ func (t *Translator) translateSecret(tctx *provider.TranslateContext, listener g
 	if tctx.Secrets == nil {
 		return nil, nil
 	}
-	if listener.TLS.CertificateRefs == nil {
-		return nil, fmt.Errorf("no certificateRefs found in listener %s", listener.Name)
-	}
 	sslObjs := make([]*adctypes.SSL, 0)
 	switch *listener.TLS.Mode {
 	case gatewayv1.TLSModeTerminate:
+		if listener.TLS.CertificateRefs == nil {
+			return nil, fmt.Errorf("no certificateRefs found in listener %s", listener.Name)
+		}
 		// frontendValidation configures downstream mTLS: clients must present a
 		// certificate signed by one of the referenced CAs during the TLS handshake.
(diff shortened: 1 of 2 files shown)


AlinsRan:
Thanks for the analysis — it's accurate: `translateSecret()` did reject Passthrough listeners before reaching the `TLSModePassthrough` case, and since `TranslateGateway` returns early on that error, the GatewayProxy global-rules/plugin-metadata aggregation was skipped as well.

This has since been fixed on master by #2804 (commit `2a8d507e`), which returns early for Passthrough before the `certificateRefs` nil check and additionally defaults a nil `listener.TLS.Mode` to Terminate. Master also already carries an equivalent `TestTranslateSecret_Passthrough`. Rebasing this PR onto master would leave an empty diff, so I'm closing it — sorry the timing worked out that way, and thanks for the report.

PR #2804, opened 2026-07-15
feat: support Gateway API 1.6.0

AlinsRan:
### Type of change:

- [x] New feature provided
- [x] CI/CD or Tests

Upgrade `sigs.k8s.io/gateway-api` from `v1.3.0` to `v1.6.0`, so the controller can run against Gateway API 1.6 and we can submit conformance reports for the current releases.

The bump pulls in `k8s.io` v0.36 and `controller-runtime` v0.24 (both require Go 1.26), so the `go` directive and CI Go version are bumped to 1.26.

**Gateway API v1.6 API changes handled:**
- Frontend mTLS validation moved from the per-listener TLS config (`listener.tls.frontendValidation`) to the Gateway level (`spec.tls.frontend`). The validation that applies to each HTTPS listener is now resolved by port (`perPort` overrides `default`) in the gateway translator, indexer, and status/reconcile paths.
- `listener.TLS` type renamed `GatewayTLSConfig` → `ListenerTLSConfig`.
- `HTTPCORSFilter.AllowCredentials` is now `*bool`.
- Policy status/target types and `Policy*` condition/reason constants graduated to `apis/v1`; usages switched off `v1alpha2`.
- `PortNumber` is now an alias of `int32`; redundant conversions dropped.

**controller-runtime v0.24 changes handled:**
- Webhook builder migrated to the generic `NewWebhookManagedBy[runtime.Object](mgr, obj)` + typed `Validator[T]` API.

**Conformance:**
- Suite rewritten for the v1.6 conformance-module API (`RunConformanceWithOptions`, report via `ReportOutputPath`).

**Tooling:**
- `golangci-lint` is built from source with the local toolchain and the lint target Go version is pinned, since the released binary cannot type-check a Go 1.26 module.

### Pre-submission checklist:

- [x] Have you added corresponding test cases? (existing unit + conformance suites updated; conformance run validates behavior)
- [x] Is this PR backward compatible? Downstream mTLS users must move `frontendValidation` from the listener to `gateway.spec.tls.frontend`, following the upstream Gateway API 1.6 change.


internal/controller/utils.go
```diff
@@ -836,51 +916,53 @@ func getListenerStatus(
...
+				Conditions: []metav1.Condition{
+					conditionProgrammed,
+					conditionAccepted,
+					conditionConflicted,
+					conditionResolvedRefs,
+				},
+				SupportedKinds: supportedKinds,
+				AttachedRoutes: attachedRoutes,
+			}))
+			continue
```
nic-6443, line 989: [P1] This only stops listener-status construction. `ParseRouteParentRefs` still matches these listeners from `Gateway.spec`, and the route reconcilers call `Provider.Update` without consulting listener status, so a TLSRoute can still be programmed even though the listeners report `Accepted=False` and `Programmed=False`; `AttachedRoutes` also remains nonzero. Please make listeners on a conflicting TLS-mode port ineligible during route attachment/translation and report no attached routes so the conflict actually prevents traffic.

internal/adc/translator/gateway.go
```diff
@@ -134,26 +146,43 @@ func (t *Translator) translateSecret(tctx *provider.TranslateContext, listener g
...
 
-// translateFrontendValidation builds the downstream mTLS client configuration from a
-// listener's frontendValidation. The referenced CA certificates (ConfigMap, key `ca.crt`)
-// are bundled into a single trust anchor used to validate client certificates.
+// frontendTLSValidation resolves the Gateway-level frontend TLS client-cert
+// validation that applies to the given HTTPS listener. In Gateway API v1.6,
+// frontendValidation moved from the per-listener TLS config to the Gateway-level
+// spec.tls.frontend: Default applies to all HTTPS listeners, and a PerPort entry
+// overrides it for listeners on the matching port.
+func frontendTLSValidation(obj *gatewayv1.Gateway, listener gatewayv1.Listener) *gatewayv1.FrontendTLSValidation {
```
nic-6443: [P2] `spec.tls.frontend.default` and `perPort` apply only to HTTPS listeners in Gateway API v1.6. This helper resolves them for every protocol, while `translateSecret` invokes it for a `TLS`/`Terminate` listener too. A mixed HTTPS + TLS Gateway can therefore attach client-certificate verification to a TLSRoute listener, or fail that listener because of an unrelated CA reference. Please return nil unless `listener.Protocol == HTTPSProtocolType` and apply the same guard in status validation.

internal/adc/translator/gateway.go
```diff
@@ -134,26 +146,43 @@ func (t *Translator) translateSecret(tctx *provider.TranslateContext, listener g
...
+	return frontend.Default.Validation
+}
+
+// translateFrontendValidation builds the downstream mTLS client configuration from the
+// Gateway's frontendValidation that applies to the listener. The referenced CA
+// certificates (ConfigMap, key `ca.crt`) are bundled into a single trust anchor used
+// to validate client certificates.
 func (t *Translator) translateFrontendValidation(tctx *provider.TranslateContext, listener gatewayv1.Listener, obj *gatewayv1.Gateway) (*adctypes.ClientClass, error) {
-	if listener.TLS.FrontendValidation == nil || len(listener.TLS.FrontendValidation.CACertificateRefs) == 0 {
+	validation := frontendTLSValidation(obj, listener)
```
nic-6443: [P2] `validation.Mode` is ignored. With `AllowInsecureFallback`, this path still builds the same CA-backed `ClientClass` as `AllowValidOnly`, so APISIX rejects clients that omit a certificate or fail verification, which is the opposite of the requested mode. Please implement fallback semantics, or reject the unsupported mode instead of silently programming strict mTLS.

internal/controller/utils.go
```diff
@@ -951,9 +1033,10 @@ func getListenerStatus(
 			}
 
 			// frontendValidation (downstream mTLS) only applies to Terminate listeners.
-			if listener.TLS.FrontendValidation != nil &&
+			// In Gateway API v1.6 it is declared at the Gateway level (spec.tls.frontend).
+			if validation := frontendTLSValidationForListener(gateway, listener); validation != nil &&
 				(listener.TLS.Mode == nil || *listener.TLS.Mode == gatewayv1.TLSModeTerminate) {
-				validateListenerFrontendValidation(ctx, mrgc, gateway, listener.TLS.FrontendValidation, &conditionResolvedRefs, &conditionProgrammed)
+				validateListenerFrontendValidation(ctx, mrgc, gateway, validation, &conditionResolvedRefs, &conditionProgrammed)
```
nic-6443: [P2] The validator only mutates `ResolvedRefs` and `Programmed`, leaving `Accepted=True`. When the sole CA ConfigMap is missing/malformed, or the Kind is unsupported, v1.6 requires `ResolvedRefs=False` with `InvalidCACertificateRef`/`InvalidCACertificateKind`, and because no valid CA remains, `Accepted=False` with `NoValidCACertificate`. Please pass/update the Accepted condition and use the CA-specific reasons.

internal/adc/translator/translator.go
```diff
@@ -66,6 +66,27 @@ func hasExplicitListenerTarget(parentRefs []gatewayv1.ParentReference) bool {
...
+// which is the correct discriminator when several listeners share a single port.
+// A server_port var adds no isolation for them and actively breaks routing: it
+// pins the route to the Gateway's declared listener port, which need not equal
+// the port APISIX actually accepts the connection on (node_listen), turning
+// every request into a 404. Only hostname-less listeners rely on port-based
+// isolation, so only their ports contribute here.
+func collectServerPortMatchPorts(listeners []gatewayv1.Listener) map[int32]struct{} {
+	ports := make(map[int32]struct{})
+	for _, listener := range listeners {
+		if listener.Hostname != nil && *listener.Hostname != "" {
```
nic-6443, line 84: [P2] `tctx.Listeners` can include both a hostname listener and a hostname-less listener when one HTTPRoute/GRPCRoute has multiple explicit ParentRefs. Dropping the hostname listener here leaves only the other port, but this set is later installed as one global `server_port` predicate on every generated APISIX route, so requests accepted through the hostname listener can never match. Please preserve all targeted ports whenever a port predicate is emitted, or split routes per listener.

internal/controller/utils.go
```diff
@@ -383,6 +383,10 @@ func ParseRouteParentRefs(
 			}
 
 			if ok, _ := routeMatchesListenerType(route, listener); !ok {
+				// The listener exists but its protocol cannot carry this route kind,
+				// which the spec reports as NotAllowedByListeners. A parentRef that
+				// matches no listener at all stays NoMatchingParent.
+				reason = gatewayv1.RouteReasonNotAllowedByListeners
```
nic-6443: [P2] For a ParentRef without `sectionName`, this assignment makes the final reason depend on listener order: a protocol-compatible listener with a hostname mismatch sets `NoMatchingListenerHostname`, then a later incompatible listener overwrites it with `NotAllowedByListeners`; reversing the list gives another result. Please aggregate candidate outcomes and select the reason after evaluating all listeners, preserving the hostname-mismatch reason when a compatible listener otherwise allows the Route.

internal/controller/utils.go
AlinsRan: Fixed. TLS-passthrough and conflicting-tls-mode listeners are now treated as non-programmable at all three attachment points, via `listenerNotProgrammable`: `ParseRouteParentRefs` no longer matches them (the parentRef gets `NotAllowedByListeners`), `getAttachedRoutesForListener` returns 0, and `getListenerStatus` reports `Accepted=False`. So a route can no longer be programmed through a listener that reports `Accepted=False`, and `AttachedRoutes` is 0. (97b3875f)

internal/adc/translator/gateway.go
AlinsRan: Fixed. Resolution is consolidated into `internaltypes.FrontendTLSValidationForListener`, which returns nil unless `listener.Protocol == HTTPSProtocolType`. Both `translateSecret` and the status validation call the same helper, so a TLS/Terminate listener never picks up frontendValidation. (51870e7c)

internal/adc/translator/gateway.go
AlinsRan: Fixed by rejecting the unsupported mode instead of silently programming strict mTLS. AllowInsecureFallback is an Extended feature that APISIX cannot express (per-SSL verification is all-or-nothing — there is no way to request a client cert without enforcing it). `translateFrontendValidation` now returns a sentinel so `translateSecret` emits no SSL for that listener, and the listener is reported `Accepted=False/UnsupportedValue` + `Programmed=False`. It is contained to the single listener — other listeners on the Gateway are unaffected (covered by `TestTranslateGateway_InsecureFallbackIsContained`). (ced6eea1)

internal/controller/utils.go
AlinsRan: Fixed. `validateListenerFrontendValidation` now also takes the `Accepted` condition: each bad ref sets `ResolvedRefs=False` with `InvalidCACertificateRef` / `InvalidCACertificateKind` (and `RefNotPermitted` for cross-namespace), and when no CA ref resolves it sets `Accepted=False/NoValidCACertificate`. (659b192a)

Note: this fixes the status conditions. The data-plane path (`processListenerConfig`) does not yet gate cross-namespace CA refs by ReferenceGrant, which is symmetric with the existing server `certificateRefs` handling — I'll follow up on aligning both rather than only the CA side.

internal/adc/translator/translator.go
AlinsRan: Addressed with a different approach — flagging for your take rather than resolving. Instead of emitting a multi-port predicate, hostname-carrying listeners are now excluded from the `server_port` set entirely (they are isolated by `service.hosts`), and a `listener_port_match_mode` knob (default off) gates port-predicate injection for the hostname-less case. So a request accepted through the hostname listener is no longer 404'd by a foreign port predicate. Splitting routes per listener would also work but is heavier — happy to go that way if you'd prefer. (a5e18fff)

internal/controller/utils.go
AlinsRan: Fixed. `ParseRouteParentRefs` now aggregates `hostnameMismatch` / `notAllowed` across all listeners and selects the reason after the loop, preferring `NoMatchingListenerHostname` when a compatible listener otherwise allows the Route — so the outcome no longer depends on listener order. (659b192a)

internal/controller/utils.go
AlinsRan: Follow-up: the data-plane gap is now closed too. `processListenerConfig` gates both server `certificateRefs` and frontendValidation CA refs with `checkReferenceGrant`, so an unauthorized cross-namespace cert/CA is no longer loaded into the data plane — status and what actually gets programmed now agree. Covered by `TestProcessListenerConfig_CrossNamespaceReferenceGrant`. (114ae153)

internal/controller/utils.go
AlinsRan: Correction: I reverted the *passthrough* half of this. Marking plain TLS Passthrough listeners as non-programmable (Accepted=False/UnsupportedValue, empty SupportedKinds) regressed three CORE GATEWAY-TLS conformance tests — they run against Passthrough listeners and require Accepted=True with SupportedKinds=[TLSRoute] (`TLSRouteListenerPassthroughSupportedKinds`, `TLSRouteInvalidNoMatchingListenerHostname`, `TLSRouteInvalidReferenceGrant`). Your ask was specifically about listeners on a *conflicting* tls.mode port, and that handling (`portsWithConflictingTLSMode` → route-ineligible + AttachedRoutes=0 + Conflicted) stays. Passthrough listeners are back to being treated as normal TLS listeners. (5bc8a0e6)

internal/adc/translator/translator.go
AlinsRan: Correction + actual fix. My earlier replies here were wrong on two counts: this scenario was **not** actually fixed, and the default mode is `auto`, not `off` — so `auto`/`explicit` both emit `server_port == 80` and drop the traffic arriving on the hostname listener's port, exactly as you described.

Fixed properly in 9b4219d4 by splitting the two concerns:
- **Decision** (whether to inject at all) still uses hostname-less listener ports, so a route bound only to hostname listeners injects nothing and `HTTPRouteListenerHostnameMatching` stays green.
- **Predicate value**, once injecting, now lists *every* targeted listener port (`allListenerPorts`), so a route attached to both a hostname-less and a hostname listener matches on both — `server_port in [80, 8080]` instead of `== 80`.

This is your option (a). Splitting routes per listener (b) remains the cleaner long-term model for real `GatewayHTTPListenerIsolation` support, but (a) removes the traffic-drop without that surgery. Test `TestTranslateHTTPRouteServerPortVarsByMode` updated — the mixed-listener case previously asserted the buggy single-port predicate.

nic-6443: [P1] The mixed-listener predicate is fixed, but `auto` remains the default while hostname-less listeners still drive injection. With a Gateway listener on port 80, APISIX `node_listen` on 9080, and a normal explicit ParentRef, this emits `server_port == 80`; the request arrives with 9080 and every route misses. The helper comment itself acknowledges that declared and accepted ports may differ. Please make this behavior opt-in with a default of `off`, or derive the actual data-plane listen port.

AlinsRan: You are right, and `explicit` is no better than `auto` here — a `sectionName` parentRef is the normal way to bind a route to a listener, so both modes take the same failure path. Deriving the data-plane listen port is not something the controller can do reliably (it cannot see `node_listen`), so the mode is now opt-in.

Fixed in 61c2bd58:
- default is `off`, and an unset or unrecognised value resolves to `off` rather than `auto`, so nothing starts injecting predicates behind the operator's back;
- docs updated — the troubleshoot entry no longer describes `off` as a workaround for the default, and the config reference/sample explain why `server_port` is not the declared listener port;
- the e2e manifest opts in explicitly, since its Gateway listeners declare APISIX's own node_listen ports, so the port-based routing specs still exercise the feature.

Same change in api7/api7-ingress-controller#419.

nic-6443 (approved):

shreemaan-abhishek (approved):

PR #2805, opened 2026-07-17
fix: enforce ReferenceGrant on cross-namespace Consumer SecretRef

shreemaan-abhishek:
### What this fixes

A `v1alpha1` `Consumer` credential's `secretRef` carries a caller-settable `namespace`:

```go
type SecretReference struct {
    Name      string  `json:"name"`
    Namespace *string `json:"namespace,omitempty"`
}
```

`ConsumerReconciler.processSpec` honored that namespace and read the Secret with the controller's cluster-wide `secrets` RBAC, with **no `ReferenceGrant` check**. A user who can only `create consumers` in namespace A could therefore name `namespace-B/some-secret` and have the controller resolve it and bind that foreign namespace's auth material (key-auth keys, basic-auth passwords, JWT secrets, HMAC keys) into a Consumer identity they own. That is a cross-namespace confused-deputy: the caller never had `get secrets` in namespace B, but the controller does.

The route controllers (`httproute`, `grpcroute`, `tcproute`, `udproute`, `tlsroute`) and the Gateway listener CA refs already gate cross-namespace references through `checkReferenceGrant`. The Consumer path was the one auth pathway that did not. Note the `v2` `ApisixConsumer` API is structurally namespace-locked (`LocalObjectReference`, no namespace field), so only `v1alpha1` was affected.

### The fix

Enforce `checkReferenceGrant` before dereferencing a cross-namespace Consumer `secretRef`, exactly as the route controllers do. Same-namespace refs are unaffected (the check short-circuits). When no grant permits the reference, the reconcile refuses it and surfaces the error on the Consumer status, so the foreign Secret is never loaded or bound.

### Tests

`internal/controller/consumer_controller_test.go` locks the invariant:

- cross-namespace `secretRef` without a `ReferenceGrant` -> refused, secret not loaded
- cross-namespace `secretRef` with a matching `ReferenceGrant` -> permitted
- same-namespace `secretRef` -> permitted, no grant needed

```
go test ./internal/controller/ -run TestProcessSpec
```

### DIFF
--- a/internal/controller/consumer_controller.go
+++ b/internal/controller/consumer_controller.go
@@ -35,6 +35,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	gatewayv1 "sigs.k8s.io/gateway-api/apis/v1"
+	"sigs.k8s.io/gateway-api/apis/v1beta1"
 
 	"github.com/apache/apisix-ingress-controller/api/v1alpha1"
 	"github.com/apache/apisix-ingress-controller/internal/controller/config"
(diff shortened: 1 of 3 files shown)


internal/controller/consumer_controller.go
```diff
@@ -253,6 +254,27 @@ func (r *ConsumerReconciler) processSpec(ctx context.Context, tctx *provider.Tra
 		if credential.SecretRef.Namespace != nil {
 			ns = *credential.SecretRef.Namespace
 		}
+		// A cross-namespace SecretRef needs a ReferenceGrant, same as routes.
+		secretNS := gatewayv1.Namespace(ns)
+		if permitted := checkReferenceGrant(ctx,
```
nic-6443, line 259: Checked this out locally and it holds up. Nice that the single change point covers admission too — `PrepareConsumerForValidation` calls `processSpec`, so the webhook gets the same gate for free, which is why the webhook test passes without touching webhook code. I reverted just this hunk and both the controller and webhook tests fail, so the tests really do pin the invariant.

One thing worth a release note: `checkReferenceGrant` returns false for a cross-namespace ref when `enableReferenceGrant` is off, and that flag is set from whether the ReferenceGrant CRD is installed (`manager/run.go`). So on a cluster without that CRD this goes from "cross-namespace secretRef works" to "always refused", not "unchecked". That's the right fail-closed direction and matches what the route controllers already do, but anyone relying on cross-namespace Consumer secretRef today will need a grant (and the CRD) after upgrading.

nic-6443 (approved):

AlinsRan (approved):

PR #2807, opened 2026-07-17
fix: reject ambiguous inline key-auth config that bypasses duplicate check

shreemaan-abhishek:
## What this PR does

The `Consumer` validating webhook extracts the key-auth `key` from inline credential config with Go's `encoding/json` struct decoder. Given a duplicate-key object like `{"key":123,"key":"K"}`, that decoder returns a type-mismatch error, so `extractCredentialKey` logs and **skips the duplicate check entirely** (returns `"", nil`). Downstream `cjson` (last-wins, no type abort) resolves the same raw bytes to a live key `"K"`.

Net effect: the webhook's only hard-deny credential check is evadable by any Consumer creator who knows a colliding key. They submit an inline config the webhook can't read but ADC/cjson can, and a duplicate active key-auth credential lands, causing consumer-identity confusion downstream. Check-bypass only (attacker must already know the key), calibrated Low.

## Fix

Parse inline key-auth config aligned with the downstream `cjson` semantics (exact-case, string-valued, last-wins) via a token-level walk, and reject the shapes that cause the divergence instead of silently skipping:

- **Duplicate `key` members** → hard admission error.
- **Non-string `key`** (number/bool/object/array) → hard admission error.
- Genuinely malformed JSON that `cjson` would also reject → still lenient (`"", nil`), so existing consumers with broken config are not suddenly denied.
- Exact-case matching means `{"Key":"K"}` is ignored (matches cjson, which only reads lowercase `key`), removing a false-positive divergence in the other direction.

## Tests

- `TestConsumerValidator_DenyDuplicateKeyAuthCredential_ParserDivergence` drives the `{"key":123,"key":"K"}` PoC end-to-end through `ValidateCreate` and asserts admission is denied.
- `TestParseInlineKeyAuthKey` table-tests the parser: valid key, duplicate members, number-then-string PoC, non-string, object value, null, missing key, exact-case, malformed, non-object.

```
go test ./internal/webhook/v1/...   # pass
go vet ./internal/webhook/...        # clean
```

Fixes FINDING-015 (api7/rfcs#149).

### DIFF
--- a/internal/webhook/v1/consumer_webhook.go
+++ b/internal/webhook/v1/consumer_webhook.go
@@ -244,6 +244,14 @@ func (v *ConsumerCustomValidator) extractCredentialKey(ctx context.Context, cons
 // check. Genuinely malformed JSON that cjson also rejects returns ("", nil) so
 // existing consumers with broken config are skipped, not suddenly denied.
 func parseInlineKeyAuthKey(raw []byte) (string, error) {
+	// cjson rejects malformed input (truncation, trailing data, multiple
+	// top-level values); mirror that by skipping anything that is not exactly
+	// one well-formed JSON value. Duplicate keys stay valid here and are caught
+	// by the token walk below.
+	if !json.Valid(raw) {
+		return "", nil
+	}
+
 	dec := json.NewDecoder(bytes.NewReader(raw))
 
 	// Top level must be an object, else there is no usable key.
--- a/internal/webhook/v1/consumer_webhook.go
+++ b/internal/webhook/v1/consumer_webhook.go
@@ -16,6 +16,7 @@
 package v1
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
@@ -227,15 +228,101 @@ func (v *ConsumerCustomValidator) extractCredentialKey(ctx context.Context, cons
 		return "", nil
 	}
 
-	var cfg struct {
-		Key string `json:"key"`
+	key, err := parseInlineKeyAuthKey(credential.Config.Raw)
+	if err != nil {
+		return "", fmt.Errorf("invalid key-auth credential config for Consumer %s/%s: %w",
+			consumer.Namespace, consumer.Name, err)
 	}
-	if err := json.Unmarshal(credential.Config.Raw, &cfg); err != nil {
-		// Malformed JSON is not a hard error: skip duplicate detection for this
-		// credential so existing consumers with bad config are not suddenly denied.
-		consumerLog.V(1).Info("skipping duplicate key-auth check: malformed credential config",
-			"consumer", consumer.Name, "error", err)
+	return key, nil
+}
+
+// parseInlineKeyAuthKey extracts the key-auth "key" from an inline credential
+// config the same way downstream cjson does: exact-case, string-valued,
+// last-wins. Ambiguous configs that Go's struct decoder would silently reject
+// while cjson still resolves to a live key (duplicate "key" members, or a
+// non-string "key") are returned as errors so they can't bypass the duplicate
+// check. Genuinely malformed JSON that cjson also rejects returns ("", nil) so
+// existing consumers with broken config are skipped, not suddenly denied.
+func parseInlineKeyAuthKey(raw []byte) (string, error) {
+	dec := json.NewDecoder(bytes.NewReader(raw))
+
+	// Top level must be an object, else there is no usable key.
+	tok, err := dec.Token()
+	if err != nil {
 		return "", nil
 	}
-	return cfg.Key, nil
+	if delim, ok := tok.(json.Delim); !ok || delim != '{' {
+		return "", nil
+	}
+
+	var (
+		key      string
+		keyCount int
+	)
+	for dec.More() {
+		nameTok, err := dec.Token()
+		if err != nil {
+			return "", nil
+		}
+		name, ok := nameTok.(string)
+		if !ok {
+			return "", nil
+		}
+		if name != "key" {
+			if err := skipJSONValue(dec); err != nil {
+				return "", nil
+			}
+			continue
+		}
+
+		keyCount++
+		valTok, err := dec.Token()
+		if err != nil {
+			return "", nil
+		}
+		switch val := valTok.(type) {
+		case string:
+			key = val
+		case nil:
+			// null key: no usable value, but still counts for dup detection.
+		default:
+			// number/bool/object/array: cjson would deliver a value here while
+			// Go's struct decoder errors and skips. Reject instead.
+			return "", fmt.Errorf("key-auth credential \"key\" must be a string")
+		}
+	}
+
+	if keyCount > 1 {
+		return "", fmt.Errorf("key-auth credential config has duplicate \"key\" members")
+	}
+	return key, nil
+}
+
+// skipJSONValue consumes a single JSON value (scalar or a whole object/array)
+// from the decoder so the token stream stays aligned.
+func skipJSONValue(dec *json.Decoder) error {
+	tok, err := dec.Token()
+	if err != nil {
+		return err
+	}
+	delim, ok := tok.(json.Delim)
+	if !ok || (delim != '{' && delim != '[') {
+		return nil
+	}
+	depth := 1
+	for depth > 0 {
+		t, err := dec.Token()
+		if err != nil {
+			return err
+		}
+		if d, ok := t.(json.Delim); ok {
+			switch d {
+			case '{', '[':
+				depth++
+			case '}', ']':
+				depth--
+			}
+		}
+	}
+	return nil
 }
--- a/internal/webhook/v1/consumer_webhook_test.go
+++ b/internal/webhook/v1/consumer_webhook_test.go
@@ -192,3 +192,67 @@ func TestConsumerValidator_DenyDuplicateKeyAuthCredential(t *testing.T) {
 	require.Contains(t, err.Error(), `duplicate key-auth credential key "shared-key"`)
 	require.Contains(t, err.Error(), "default/existing")
 }
+
+// A duplicate-key inline config ({"key":123,"key":"K"}) is unreadable to Go's
+// struct decoder but resolves to "K" downstream via cjson. The webhook must
+// reject it instead of silently skipping the duplicate check.
+func TestConsumerValidator_DenyDuplicateKeyAuthCredential_ParserDivergence(t *testing.T) {
+	existing := &apisixv1alpha1.Consumer{
+		ObjectMeta: metav1.ObjectMeta{Name: "existing", Namespace: "default"},
+		Spec: apisixv1alpha1.ConsumerSpec{
+			GatewayRef: apisixv1alpha1.GatewayRef{Name: "test-gateway"},
+			Credentials: []apisixv1alpha1.Credential{{
+				Type:   "key-auth",
+				Config: apiextensionsv1.JSON{Raw: []byte(`{"key":"victims-key"}`)},
+			}},
+		},
+	}
+	consumer := &apisixv1alpha1.Consumer{
+		ObjectMeta: metav1.ObjectMeta{Name: "demo", Namespace: "default"},
+		Spec: apisixv1alpha1.ConsumerSpec{
+			GatewayRef: apisixv1alpha1.GatewayRef{Name: "test-gateway"},
+			Credentials: []apisixv1alpha1.Credential{{
+				Type:   "key-auth",
+				Config: apiextensionsv1.JSON{Raw: []byte(`{"key":123,"key":"victims-key"}`)},
+			}},
+		},
+	}
+
+	validator := buildConsumerValidator(t, existing)
+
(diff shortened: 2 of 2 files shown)


shreemaan-abhishek:
Synced to the enterprise ingress controller in api7/api7-ingress-controller#434 (both must merge to complete the sync).

shreemaan-abhishek:
Good catch. \`parseInlineKeyAuthKey\` now guards with \`json.Valid\` up front, so truncated input, trailing data, and multiple top-level values are treated as malformed and skipped (matching cjson, which rejects them too). Duplicate keys remain valid JSON and are still caught by the token walk. Added table cases for the truncated-object, multi-value, and trailing-garbage shapes. Same fix synced to api7/api7-ingress-controller#434.

internal/webhook/v1/consumer_webhook.go
```diff
@@ -227,15 +228,109 @@ func (v *ConsumerCustomValidator) extractCredentialKey(ctx context.Context, cons
...
+			// null key: no usable value, but still counts for dup detection.
+		default:
+			// number/bool/object/array: cjson would deliver a value here while
+			// Go's struct decoder errors and skips. Reject instead.
+			return "", fmt.Errorf("key-auth credential \"key\" must be a string")
+		}
+	}
+
+	if keyCount > 1 {
+		return "", fmt.Errorf("key-auth credential config has duplicate \"key\" members")
```
nic-6443, line 304: I don't think this branch can fire in a real cluster. The apiserver collapses duplicate JSON object keys (last-wins) while decoding the request — before admission webhooks are called and before storage — so `credential.Config.Raw` never holds two `key` members.

I checked at the raw-byte level rather than trusting a round-trip through a JSON parser (which would collapse duplicates itself and hide the answer): registered a probe validating webhook on `consumers`, dumped the unparsed AdmissionReview body, and POSTed the PoC through the raw REST endpoint.

```
$ kubectl create --raw ".../consumers" -f -   # {"key":123,"key":"victims-key"}
Warning: duplicate field "spec.credentials[0].config.key"

# raw bytes the webhook received:
{"credentials":[{"config":{"key":"victims-key"},"name":"c1","type":"key-auth"}],...
#   occurrences of "key": -> 1     contains "key":123 -> False
```

The persisted object is that same normalized form, so the controller and cjson downstream see it too. Which means the old struct decoder parses the PoC fine and the duplicate check runs normally — the webhook and cjson never disagree, so there's nothing to bypass.

Do you have a PoC that reaches the webhook with the duplicate intact? If it was built by calling `ValidateCreate` directly in a test with hand-written `Raw` bytes, that would explain the gap — those bytes can't come out of the apiserver.

internal/webhook/v1/consumer_webhook.go
```diff
@@ -227,15 +228,109 @@ func (v *ConsumerCustomValidator) extractCredentialKey(ctx context.Context, cons
 		return "", nil
 	}
 
-	var cfg struct {
-		Key string `json:"key"`
+	key, err := parseInlineKeyAuthKey(credential.Config.Raw)
+	if err != nil {
+		return "", fmt.Errorf("invalid key-auth credential config for Consumer %s/%s: %w",
```
nic-6443, line 233: This part I think is a real regression: it turns a per-credential skip into a hard admission failure, and the caller applies that to *other* objects too.

`validateDuplicateKeyAuthCredentials` loops over every existing Consumer sharing the gateway and does `extractKeyAuthKeys(existing)` -> `return err`. So one stored Consumer with e.g. `{"key":123}` now denies create/update of every unrelated Consumer on that gateway, with an error naming a different object.

That state is reachable, unlike the duplicate-key case: `config` is preserve-unknown-fields so the apiserver stores `{"key":123}` verbatim, and the webhook ships `failurePolicy: Ignore`, so anything created while the controller is down lands unvalidated.

```go
// existing: {"key":123}   newcomer: {"key":"my-own-unique-key"}, same gateway, no collision
validator := buildConsumerValidator(t, existing)
_, err := validator.ValidateCreate(context.Background(), newcomer)
require.NoError(t, err)
```

Passes on master, fails here with `invalid key-auth credential config for Consumer default/legacy: key-auth credential "key" must be a string`.

Could the existing-consumers loop stay best-effort — skip a credential it can't read, as today — and hard-deny only on the incoming object? That keeps the stricter validation where it's actionable without letting one legacy object wedge a whole gateway.

nic-6443 (changes requested):

internal/webhook/v1/consumer_webhook.go
AlinsRan, line 233: Blocking: this error propagates out of `extractKeyAuthKeys(ctx, existing)` in the loop over *other* Consumers (`validateDuplicateKeyAuthCredentials`), so one pre-existing object with a bad inline config denies admission for every other Consumer on the same gateway.

Reproduced against this branch — a `legacy` Consumer with `{"key":123}` already in the cluster, then an unrelated create with a perfectly valid, non-colliding key:

```
ValidateCreate(innocent) err =
  invalid key-auth credential config for Consumer default/legacy: key-auth credential "key" must be a string
```

The error even names someone else's object. The code this replaces called this out explicitly ("existing consumers with bad config are not suddenly denied").

If you keep any hard-deny path, it has to apply only to the object under admission; when walking other Consumers, a parse failure should degrade to skip + log.

internal/webhook/v1/consumer_webhook.go
```diff
@@ -227,15 +228,109 @@ func (v *ConsumerCustomValidator) extractCredentialKey(ctx context.Context, cons
...
+			consumer.Namespace, consumer.Name, err)
+	}
+	return key, nil
+}
+
+// parseInlineKeyAuthKey extracts the key-auth "key" from an inline credential
+// config the same way downstream cjson does: exact-case, string-valued,
```
AlinsRan, line 240: `the same way downstream cjson does` — the downstream here is not cjson. It's `json.Unmarshal(credentialSpec.Config.Raw, &authConfig)` into a `map[string]any` at `internal/adc/translator/consumer.go:65`. Worth correcting the comment, because it's what the whole duplicate-member branch is justified by.

And since duplicate members can't survive the API server's unstructured decode (see the top-level comment), `keyCount > 1` is dead in practice for anything created through the Kubernetes API.

internal/webhook/v1/consumer_webhook.go
```diff
@@ -227,15 +228,109 @@ func (v *ConsumerCustomValidator) extractCredentialKey(ctx context.Context, cons
...
+// while cjson still resolves to a live key (duplicate "key" members, or a
+// non-string "key") are returned as errors so they can't bypass the duplicate
+// check. Genuinely malformed JSON that cjson also rejects returns ("", nil) so
+// existing consumers with broken config are skipped, not suddenly denied.
+func parseInlineKeyAuthKey(raw []byte) (string, error) {
+	// cjson rejects malformed input (truncation, trailing data, multiple
+	// top-level values); mirror that by skipping anything that is not exactly
+	// one well-formed JSON value. Duplicate keys stay valid here and are caught
+	// by the token walk below.
+	if !json.Valid(raw) {
```
AlinsRan, line 251: If the goal is just to match the translator's semantics (exact-case, last-wins), `map[string]json.RawMessage` *is* that semantics — no token walk needed:

```go
var cfg map[string]json.RawMessage
if err := json.Unmarshal(raw, &cfg); err != nil {
	return "", nil // downstream fails the same way; stay lenient
}
v, ok := cfg["key"] // exact-case, last-wins, matches the translator
if !ok {
	return "", nil
}
var s string
if err := json.Unmarshal(v, &s); err != nil {
	return "", nil // non-string: the key-auth schema rejects it downstream anyway
}
return s, nil
```

That's ~10 lines instead of ~90, fixes the `{"Key":"K"}` false positive, and adds no new denial paths. Note also that `json.Valid` here parses the input a second time — `dec.Token()` failing already covers malformed input.

Separately: the `consumerLog.V(1).Info("skipping duplicate key-auth check: ...")` line was dropped, so the lenient-skip path is now completely silent. Worth keeping.

internal/webhook/v1/consumer_webhook.go
```diff
@@ -227,15 +228,109 @@ func (v *ConsumerCustomValidator) extractCredentialKey(ctx context.Context, cons
...
+
+	if keyCount > 1 {
+		return "", fmt.Errorf("key-auth credential config has duplicate \"key\" members")
+	}
+	return key, nil
+}
+
+// skipJSONValue consumes a single JSON value (scalar or a whole object/array)
+// from the decoder so the token stream stays aligned.
+func skipJSONValue(dec *json.Decoder) error {
```
AlinsRan, line 311: If the token walk stays, this whole function is `dec.Decode(&json.RawMessage{})` — `Decode` consumes exactly one value and keeps the stream aligned.

internal/webhook/v1/consumer_webhook_test.go
```diff
@@ -192,3 +192,70 @@ func TestConsumerValidator_DenyDuplicateKeyAuthCredential(t *testing.T) {
...
+			}},
+		},
+	}
+	consumer := &apisixv1alpha1.Consumer{
+		ObjectMeta: metav1.ObjectMeta{Name: "demo", Namespace: "default"},
+		Spec: apisixv1alpha1.ConsumerSpec{
+			GatewayRef: apisixv1alpha1.GatewayRef{Name: "test-gateway"},
+			Credentials: []apisixv1alpha1.Credential{{
+				Type:   "key-auth",
+				Config: apiextensionsv1.JSON{Raw: []byte(`{"key":123,"key":"victims-key"}`)},
```
AlinsRan, line 216: This is the crux of the reachability problem: `Config.Raw` is set directly in Go, so the bytes never pass through the API server's unstructured decode that would have collapsed the duplicate to `{"key":"victims-key"}`. A Consumer created via `kubectl`/client-go can't reach the validator in this state, so this doesn't demonstrate an exploitable path.

What's missing is a case for the regression this introduces: a pre-existing Consumer with a broken inline config, plus an unrelated new Consumer with a valid key, asserting the new one is still admitted. That currently fails.

AlinsRan (changes requested):
I don't think this should merge in its current shape — two problems, one about the premise and one about a regression I reproduced.

**The duplicate-key bypass isn't reachable through the API server.** CRs are decoded into `unstructured` (`map[string]interface{}`) before they're stored or sent in an AdmissionReview, and that decode collapses duplicate members last-wins. I ran the object body through the same decode path:

```
in:  {"credentials":[{"type":"key-auth","config":{"key":123,"key":"victims-key"}}]}
out: {"config":{"key":"victims-key"},"type":"key-auth"}
```

The webhook and the translator read the same normalized bytes, so there is no window where one sees `{"key":123,"key":"K"}` and the other sees `"K"`. The PoC only exists because the test constructs `Config.Raw` in Go directly.

**The fix makes one bad pre-existing object deny everyone else on the gateway.** Verified — details inline.

There *is* a real divergence in here worth fixing, just not the one described: Go's struct decoder is case-insensitive, so `{"Key":"K"}` currently registers as key `K` in the webhook while the translator's map decode keeps `"Key"` and APISIX never sees a `key`. That's a false-positive denial, and exact-case matching fixes it. I'd narrow the PR to that.

Also, the description says the downstream is cjson with last-wins semantics. The actual downstream is `json.Unmarshal(credentialSpec.Config.Raw, &map[string]any{})` in `internal/adc/translator/consumer.go:65` — a Go map decoder. The raw bytes never reach APISIX's Lua layer in their original form.

shreemaan-abhishek:
Closing: the reviewer is right, and I reproduced it.

The Kubernetes apiserver collapses duplicate JSON object keys (last-wins) while decoding the request, before admission webhooks run and before storage. `config` is `x-kubernetes-preserve-unknown-fields` (an object), so it is still fully parsed and re-serialized. POSTing the PoC through the raw endpoint confirms the duplicate never survives:

```
$ kubectl create --raw '/apis/apisix.apache.org/v1alpha1/namespaces/default/consumers' -f - <<'JSON'
{"apiVersion":"apisix.apache.org/v1alpha1","kind":"Consumer","metadata":{"name":"poc","namespace":"default"},
 "spec":{"gatewayRef":{"name":"test-gateway"},
 "credentials":[{"type":"key-auth","name":"c1","config":{"key":123,"key":"victims-key"}}]}}
JSON
Warning: duplicate field "spec.credentials[0].config.key"

# stored + admitted object:
"credentials":[{"config":{"key":"victims-key"},"name":"c1","type":"key-auth"}]
```

So the webhook's `Config.Raw` is the normalized `{"key":"victims-key"}` - the original struct decoder parses it fine and the duplicate check runs; webhook and cjson never disagree. The PoC only fired because it fed hand-written `Raw` bytes directly to `ValidateCreate`, which the apiserver cannot produce.

The second review comment is also correct: the change regressed the existing-consumers loop from a per-credential skip into a gateway-wide denial when a single stored Consumer has a non-string key config. Both reasons to drop this. The underlying finding is not exploitable via the admission path. Thanks for the thorough review.

PR #2808, opened 2026-07-17
fix: redact credentials and AdminKey from controller logs

shreemaan-abhishek:
## What this PR does

Several controller log sites serialize whole, secret-bearing objects, so
credentials end up in controller stdout / log aggregation on routine paths
(a weaker boundary than `secrets/get` RBAC). This redacts them.

### Leaks fixed

- **AdminKey** (`ADCServerOpts.Token`): the ADC request body was logged verbatim
  at `V(1)` in `buildHTTPRequest`, printing the plaintext admin key.
- **TLS private keys & consumer credentials** (`adctypes.Resources`): `Task`
  logging at ERROR (`store insert/delete failed`) and `V(1)` (`updating config`)
  dumped `SSL.Certificates[].Key` and consumer credential material.
- **Inline plugin credentials**: controllers logged full `ApisixConsumer`,
  `Consumer`, `ApisixRoute`, `ApisixGlobalRule` and filtered `HTTPRoute` objects,
  whose specs can carry inline plugin config (key-auth key, hmac/jwt secrets).
- **Raw credential/plugin blobs**: the consumer translator logged the raw
  credential/plugin JSON on unmarshal errors.
- **Global-rule plugin config**: the store logged full global-rule items
  (logger-plugin tokens).

### Approach

- Add `MarshalLog` (`logr.Marshaler`) to `adctypes.Resources`, `client.Task` and
  `ADCServerRequest`. These emit identity + counts, and `[REDACTED]` for the
  token, never secret bodies. `MarshalLog` affects **logging only** - the JSON
  actually sent to the data plane is untouched (a `MarshalJSON`-based redaction
  would have corrupted the wire payload, which is why it is not used here).
- Log only `NamespacedName` / identity at the controller, provider, translator
  and store sites.

### Tests

`internal/adc/client/redaction_test.go` builds the production logger
(`zapr` + zap console encoder) into a buffer, logs a `Task` and an
`ADCServerRequest` populated with a TLS key, a consumer credential and an admin
key, and asserts none of the secrets appear in the output while identity still
does. It also asserts the request's real `Token` field is unchanged after
logging.

```
go test ./internal/adc/client/ -run MarshalLog -v
```

### DIFF
--- a/api/adc/types.go
+++ b/api/adc/types.go
@@ -61,6 +61,23 @@ type Resources struct {
 	SSLs           []*SSL           `json:"ssls,omitempty" yaml:"ssls,omitempty"`
 }
 
+// MarshalLog implements logr.Marshaler so logging Resources emits only counts,
+// never the secret-bearing bodies (SSL private keys, consumer credentials).
+// It affects logging only, not the JSON sent to the data plane.
+func (r *Resources) MarshalLog() any {
+	if r == nil {
+		return nil
+	}
+	return map[string]int{
+		"consumerGroups": len(r.ConsumerGroups),
+		"consumers":      len(r.Consumers),
+		"globalRules":    len(r.GlobalRules),
+		"pluginMetadata": len(r.PluginMetadata),
+		"services":       len(r.Services),
+		"ssls":           len(r.SSLs),
+	}
+}
+
 type GlobalRule Plugins
 
 func (g *GlobalRule) DeepCopy() GlobalRule {
--- a/internal/adc/cache/store.go
+++ b/internal/adc/cache/store.go
@@ -230,7 +230,7 @@ func (s *Store) GetResources(name string) (*adctypes.Resources, error) {
 		}
 		globalrule = adctypes.GlobalRule(merged)
 	}
-	s.log.V(1).Info("GetResources fetched global rule items", "items", globalRuleItems, "gobalrule", globalrule)
+	s.log.V(1).Info("GetResources fetched global rule items", "itemCount", len(globalRuleItems), "pluginCount", len(globalrule))
 	if meta, ok := s.pluginMetadataMap[name]; ok {
 		metadata = meta.DeepCopy()
 	}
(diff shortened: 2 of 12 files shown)


api/adc/types.go
```diff
@@ -61,6 +61,23 @@ type Resources struct {
 	SSLs           []*SSL           `json:"ssls,omitempty" yaml:"ssls,omitempty"`
 }
 
+// MarshalLog implements logr.Marshaler so logging Resources emits only counts,
+// never the secret-bearing bodies (SSL private keys, consumer credentials).
+// It affects logging only, not the JSON sent to the data plane.
+func (r *Resources) MarshalLog() any {
```
nic-6443, line 67: Mechanism checks out — I built the production logger (zapr + zap console encoder) into a buffer and confirmed `MarshalLog` is honoured at all three sites, the token comes out `[REDACTED]`, and the wire payload is untouched. Also confirmed the remaining `"config", config` / `"configs", configs` sites in `client.go` are already safe: `Config.MarshalJSON` gets picked up by zap's reflect encoder, so the AdminKey doesn't survive there either.

One sharp edge on this receiver: it's on `*Resources`, so `Resources` by value doesn't satisfy `logr.Marshaler` and silently falls back to reflection. Logging a value dumps everything:

```
log.V(1).Info("resources", "resources", *someResources)
// {"resources": {"consumers":[{"plugins":{"key-auth":{"key":"SUPER-SECRET"}}...
```

No current call site does that (`Task.Resources` is a pointer and `ADCServerRequest.MarshalLog` calls it explicitly), so nothing to fix here — but a value receiver would make it hold under refactoring, since the pointer method set includes it either way. Up to you.

Separately, and out of scope for this PR: `translator/httproute.go` still logs the fully-decoded plugin map from a PluginConfig extensionRef at V(1), and `translator/gateway.go` does the same for GatewayProxy plugins plus the raw plugin_metadata blob on unmarshal error. Same "inline plugin credentials" class you're closing here. Worth a follow-up?

nic-6443 (approved):

AlinsRan (approved):

PR #2809, opened 2026-07-19
fix: parse hmac-auth signed_headers from Secret as a header list

shreemaan-abhishek:
### What this PR does

`translateConsumerHMACAuthPlugin` read `signed_headers` from a Secret by ranging over its raw bytes:

```go
signedHeadersRaw := sec.Data["signed_headers"] // []byte
for _, b := range signedHeadersRaw {           // ranges over bytes
    signedHeaders = append(signedHeaders, string(b)) // one 1-char entry per byte
}
```

Ranging over a `[]byte` yields `(index, byte)`, so `"X-Date,Host"` became `["X","-","D","a","t","e",",","H","o","s","t"]`. The data-plane hmac-auth policy then bound those single-character names into the signature, which never match real headers, so the operator's intended headers were silently **not** enforced as part of the HMAC signature. Only the `secretRef` path was affected; the inline `Value` path already passes a `[]string`.

### Fix

- Split the `signed_headers` value on commas and trim entries.
- Surface `strconv.ParseInt` failures for `clock_skew` and `max_req_body` instead of discarding them, so a typo no longer coerces silently to a default.

### Tests

- New unit tests: `signed_headers` from a Secret yields the correct header list; an unparseable `clock_skew` is rejected.

Scope: `internal/adc/translator/apisixconsumer.go`, CRD-to-data-plane fidelity. Self-scoped to the consumer owner's own config.

### DIFF
--- a/internal/adc/translator/apisixconsumer.go
+++ b/internal/adc/translator/apisixconsumer.go
@@ -20,6 +20,7 @@ package translator
 import (
 	"fmt"
 	"strconv"
+	"strings"
 
 	"github.com/pkg/errors"
 	k8stypes "k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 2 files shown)


internal/adc/translator/apisixconsumer.go
```diff
@@ -307,15 +308,25 @@ func (t *Translator) translateConsumerHMACAuthPlugin(tctx *provider.TranslateCon
...
 		clockSkew = _hmacAuthClockSkewDefaultValue
 	}
 
+	// comma-separated header names, not raw bytes
 	signedHeadersRaw := sec.Data["signed_headers"]
-	signedHeaders := make([]string, 0, len(signedHeadersRaw))
-	for _, b := range signedHeadersRaw {
-		signedHeaders = append(signedHeaders, string(b))
+	var signedHeaders []string
+	for _, h := range strings.Split(string(signedHeadersRaw), ",") {
```
AlinsRan, line 326: New convention: the Secret path now needs comma-separated `signed_headers`, while the inline `Value` path takes a native `[]string`. Undocumented — worth a line in the CRD field doc, else newline/space-separated input silently yields one wrong header.

internal/adc/translator/apisixconsumer.go
```diff
@@ -307,15 +308,25 @@ func (t *Translator) translateConsumerHMACAuthPlugin(tctx *provider.TranslateCon
 	}
 
 	clockSkewRaw := sec.Data["clock_skew"]
-	clockSkew, _ := strconv.ParseInt(string(clockSkewRaw), 10, 64)
+	var clockSkew int64
+	if len(clockSkewRaw) > 0 {
+		var err error
+		clockSkew, err = strconv.ParseInt(string(clockSkewRaw), 10, 64)
+		if err != nil {
+			return nil, fmt.Errorf("hmac-auth: invalid clock_skew %q in secret %s/%s: %w", string(clockSkewRaw), consumerNamespace, cfg.SecretRef.Name, err)
```
AlinsRan, line 316: Semantics change for the changelog: a consumer with an unparseable `clock_skew`/`max_req_body` used to default silently and sync; now it hard-fails and stops syncing. Right direction, self-scoped to that consumer. The `len(...) > 0` guard correctly keeps the empty case on the default path.

AlinsRan (approved):
Real bug, fix is correct. The byte-range turned `signed_headers` into per-character entries, so the signature bound garbage names and the intended headers went unenforced. The inline `Value` path is unaffected (already `[]string`). Two minor notes inline.

nic-6443 (approved):

PR #2810, opened 2026-07-19
fix: detect and warn on inconsistent SSL host/certificate configuration

shreemaan-abhishek:
### Description

Improves how the controller detects and reports inconsistent TLS host/certificate configuration. The SSL conflict detector now **warns** on overlap instead of **denying** it at admission (see the note below), and its detection is made wildcard- and mTLS-aware so the warnings are accurate.

**1. Wildcard vs. exact host matching**

Hosts were compared by exact string equality, so a covering wildcard and an exact host were never compared against each other (`*.example.com` and `app.example.com` are indexed under different keys). APISIX resolves an exact SNI ahead of a covering wildcard, so two objects whose hosts overlap only through a wildcard collide at the data plane.

- add `sslutil.HostsOverlap` / `ParentWildcard` (single-label wildcard semantics)
- for an exact host, also look up its covering wildcard key
- for a wildcard host, enumerate TLS resources and filter by overlap
- compare candidate mappings with `HostsOverlap` instead of string equality

**2. mTLS client config in the conflict key**

The conflict key used only the server certificate hash, so two objects for the same host and server certificate but different mTLS client config (`spec.client`) were not flagged as overlapping.

- add `ClientConfigHash` to `HostCertMapping` (CA secret reference, depth, `skip_mtls_uri_regex`; empty when no mTLS)

**3. Overlap is reported as a warning, not a denial**

The detector previously **rejected** overlapping SSL objects at admission. That gate only sees object create/update, so it misses conflicts introduced by a certificate rotating inside its Secret, a `GatewayProxy`/`IngressClass` regrouping, a disabled webhook, or objects that predate the webhook; and it blocks edits to configurations that already coexist. APISIX admits overlapping SSL objects and resolves by SNI specificity, so overlap is now surfaced as an **admission warning** on create and update.

A reconcile-time treatment (detect on reconcile + an `OverlappingTLSConfig`-style status condition with a documented tie-break) is planned as a follow-up rather than expanding this PR.

**4. Certificate SAN coverage warning (translator)**

`TranslateApisixTls` programmed the SSL object without any signal when the certificate SANs don't cover a declared host. APISIX serves the cert regardless (the client validates), so this logs a warning listing the uncovered hosts, using the wildcard-aware `sslutil.HostCoveredBy`.

### Tests

- unit tests for `HostsOverlap` / `ParentWildcard` / `HostCoveredBy`
- detector-level tests for wildcard/exact overlap (both directions) and differing mTLS config, each with a false-positive guard
- translator tests for SAN coverage (exact, wildcard, mismatch, partial, no-SAN lenient)

### Checklist

### DIFF
--- a/internal/adc/translator/apisixtls.go
+++ b/internal/adc/translator/apisixtls.go
@@ -50,6 +50,14 @@ func (t *Translator) TranslateApisixTls(tctx *provider.TranslateContext, tls *ap
 		return nil, err
 	}
 
+	// APISIX serves the cert regardless of SAN, so this is advisory: warn when a
+	// declared host isn't covered by the cert SANs (clients may reject it).
+	if uncovered, sans := uncoveredSNIHosts(cert, tls.Spec.Hosts); len(uncovered) > 0 {
+		t.Log.Info("ApisixTls certificate does not cover all declared SNI hosts",
+			"namespace", tls.Namespace, "name", tls.Name,
+			"uncoveredHosts", uncovered, "certificateSANs", sans)
+	}
+
 	// Convert hosts to strings
 	snis := make([]string, len(tls.Spec.Hosts))
 	for i, host := range tls.Spec.Hosts {
@@ -96,4 +104,29 @@ func (t *Translator) TranslateApisixTls(tctx *provider.TranslateContext, tls *ap
 
 	result.SSL = append(result.SSL, ssl)
 	return result, nil
+}
+
+// uncoveredSNIHosts returns the declared hosts not covered by any DNS SAN in the
+// certificate (wildcard-aware), plus the cert SANs for logging. Returns nothing
+// when the cert declares no DNS SANs or can't be parsed, since coverage can't be
+// judged there.
+func uncoveredSNIHosts(cert []byte, hosts []apiv2.HostType) (uncovered []string, sans []string) {
+	sans, err := sslutils.ExtractHostsFromCertificate(cert)
+	if err != nil || len(sans) == 0 {
+		return nil, nil
+	}
+	for _, h := range hosts {
+		host := string(h)
+		covered := false
+		for _, san := range sans {
+			if sslutils.HostCoveredBy(host, san) {
+				covered = true
+				break
+			}
+		}
+		if !covered {
+			uncovered = append(uncovered, host)
+		}
+	}
+	return uncovered, sans
 }
--- a/internal/ssl/util.go
+++ b/internal/ssl/util.go
@@ -218,6 +218,25 @@ func wildcardCovers(wildcard, host string) bool {
 	return label != "" && !strings.Contains(label, ".")
 }
 
+// HostCoveredBy reports whether a certificate SAN covers an SNI host. Coverage
+// is directional: an exact SAN covers only the identical host; a wildcard SAN
+// "*.suffix" covers single-label subdomains "<label>.suffix". A wildcard host is
+// covered only by an identical wildcard SAN (an exact SAN can't serve it).
+func HostCoveredBy(host, san string) bool {
+	host = strings.ToLower(strings.TrimSpace(host))
+	san = strings.ToLower(strings.TrimSpace(san))
+	if host == "" || san == "" {
+		return false
+	}
+	if host == san {
+		return true
+	}
+	if strings.HasPrefix(san, "*.") {
+		return wildcardCovers(san, host)
+	}
+	return false
+}
+
 // ParentWildcard returns the single-label covering wildcard for an exact host,
 // e.g. "app.example.com" -> "*.example.com". Returns "" if host is empty, is
 // itself a wildcard, or has no parent label.
--- a/internal/ssl/util.go
+++ b/internal/ssl/util.go
@@ -180,6 +180,58 @@ func NormalizeHosts(hosts []string) []string {
 	return normalized
 }
 
+// HostsOverlap reports whether two SNI host patterns can both match a common
+// concrete hostname. It understands single-label wildcards ("*.example.com"
+// matches "app.example.com" but not "a.b.example.com"). Two distinct wildcards
+// never share a concrete host.
+func HostsOverlap(a, b string) bool {
+	a = strings.ToLower(strings.TrimSpace(a))
+	b = strings.ToLower(strings.TrimSpace(b))
+	if a == "" || b == "" {
+		return false
+	}
+	if a == b {
+		return true
+	}
+	aWild := strings.HasPrefix(a, "*.")
+	bWild := strings.HasPrefix(b, "*.")
+	switch {
+	case aWild && bWild:
+		return false
+	case aWild:
+		return wildcardCovers(a, b)
+	case bWild:
+		return wildcardCovers(b, a)
+	default:
+		return false
+	}
+}
+
+// wildcardCovers reports whether wildcard "*.suffix" matches the exact host,
+// i.e. host is exactly "<single-label>.suffix".
+func wildcardCovers(wildcard, host string) bool {
+	suffix := wildcard[1:] // ".example.com"
+	if !strings.HasSuffix(host, suffix) {
+		return false
+	}
+	label := host[:len(host)-len(suffix)]
+	return label != "" && !strings.Contains(label, ".")
+}
+
+// ParentWildcard returns the single-label covering wildcard for an exact host,
+// e.g. "app.example.com" -> "*.example.com". Returns "" if host is empty, is
+// itself a wildcard, or has no parent label.
+func ParentWildcard(host string) string {
+	if host == "" || strings.HasPrefix(host, "*.") {
+		return ""
+	}
+	i := strings.IndexByte(host, '.')
+	if i < 0 || i == len(host)-1 {
+		return ""
+	}
+	return "*." + host[i+1:]
+}
+
 // CertificateHash returns the SHA-256 hash of the leaf certificate contained in the PEM data.
 // The hash is calculated from the DER-encoded bytes so that formatting differences (whitespace,
 // line endings, certificate ordering) do not affect the result.
--- a/internal/webhook/v1/apisixtls_webhook.go
+++ b/internal/webhook/v1/apisixtls_webhook.go
@@ -74,16 +74,17 @@ func (v *ApisixTlsCustomValidator) ValidateCreate(ctx context.Context, obj runti
 	}
 
 	detector := sslvalidator.NewConflictDetector(v.Client)
-	conflicts := detector.DetectConflicts(ctx, tls)
-	if len(conflicts) > 0 {
-		return nil, fmt.Errorf("%s", sslvalidator.FormatConflicts(conflicts))
-	}
-
 	warnings := v.collectWarnings(ctx, tls)
+	// Overlapping SSL config is advisory: APISIX serves overlapping objects and
+	// resolves by SNI, so warn instead of denying.
+	skipADCValidation := v.initErr != nil || len(warnings) > 0
+	if conflicts := detector.DetectConflicts(ctx, tls); len(conflicts) > 0 {
+		warnings = append(warnings, sslvalidator.FormatConflicts(conflicts))
+	}
 	// Skip ADC validation when secrets are missing: the translator cannot
 	// load cert/key material, so validation would always fail. The missing-
 	// secret warnings are sufficient to inform the user.
-	if v.initErr != nil || len(warnings) > 0 {
+	if skipADCValidation {
 		return warnings, nil
 	}
 
--- a/internal/webhook/v1/ssl/conflict_detector.go
+++ b/internal/webhook/v1/ssl/conflict_detector.go
@@ -282,10 +282,15 @@ func clientConfigHash(client *apiv2.ApisixMutualTlsClientConfig) string {
 	}
 	regexes := append([]string(nil), client.SkipMTLSUriRegex...)
 	sort.Strings(regexes)
-	canonical := fmt.Sprintf("ca=%s/%s;depth=%d;skip=%s",
-		client.CASecret.Namespace, client.CASecret.Name, client.Depth,
-		strings.Join(regexes, ","))
-	sum := sha256.Sum256([]byte(canonical))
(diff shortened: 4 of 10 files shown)


internal/webhook/v1/ssl/conflict_detector.go
```diff
@@ -251,17 +258,37 @@ func (d *ConflictDetector) BuildApisixTlsMappings(ctx context.Context, tls *apiv
...
+// the trust anchor, plus depth and the skip regexes.
+func clientConfigHash(client *apiv2.ApisixMutualTlsClientConfig) string {
+	if client == nil {
+		return ""
+	}
+	regexes := append([]string(nil), client.SkipMTLSUriRegex...)
+	sort.Strings(regexes)
+	canonical := fmt.Sprintf("ca=%s/%s;depth=%d;skip=%s",
+		client.CASecret.Namespace, client.CASecret.Name, client.Depth,
+		strings.Join(regexes, ","))
```
nic-6443: [P2] This canonicalization is not injective: `{"a,b"}` and `{"a", "b"}` both produce `skip=a,b`, and commas are legal regex characters. Two different mTLS skip configurations can therefore hash identically and bypass conflict detection. Please encode each entry unambiguously, for example with length prefixes or a serialized sorted slice.

internal/webhook/v1/ssl/conflict_detector.go
```diff
@@ -92,22 +98,23 @@ func (d *ConflictDetector) DetectConflicts(ctx context.Context, obj client.Objec
 	conflicts := make([]SSLConflict, 0)
 
 	// First, check for conflicts within the new resource itself.
-	seen := make(map[string]string, len(newMappings))
+	seen := make(map[string]HostCertMapping, len(newMappings))
 	for _, mapping := range newMappings {
 		if mapping.Host == "" || mapping.CertificateHash == "" {
 			continue
```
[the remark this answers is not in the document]
nic-6443: This remains valid on the current head: `seen[mapping.Host]` still compares exact strings, while the external path uses `HostsOverlap`. A single object can therefore admit `*.example.com` and `app.example.com` with different certificate or client config. Please make the intra-resource check overlap-aware and add coverage for both Ingress TLS blocks and Gateway listeners.

internal/webhook/v1/ssl/conflict_detector.go
```diff
@@ -376,8 +433,12 @@ func (d *ConflictDetector) findExternalConflicts(ctx context.Context, obj client
 			if !ok {
 				continue
 			}
-			// same cert hash, no conflict
-			if mapping.CertificateHash == hosts[host] {
+			// Same server cert AND same mTLS client config: no conflict. A
+			// differing client config is still a conflict even when the server
+			// cert matches.
+			newMapping := newMappings[host]
+			if mapping.CertificateHash == newMapping.CertificateHash &&
```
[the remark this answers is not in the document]
nic-6443: This also remains on the current head. `ClientConfigHash` can be the only differing field, but `SSLConflict` carries no cause and `FormatConflicts` always reports a “different certificate”. Please report a generic differing TLS configuration, or carry the differing field into the diagnostic.

internal/ssl/util.go
```diff
@@ -180,6 +180,77 @@ func NormalizeHosts(hosts []string) []string {
 	return normalized
 }
 
+// HostsOverlap reports whether two SNI host patterns can both match a common
+// concrete hostname. It understands single-label wildcards ("*.example.com"
+// matches "app.example.com" but not "a.b.example.com"). Two distinct wildcards
+// never share a concrete host.
+func HostsOverlap(a, b string) bool {
```
AlinsRan, line 187: An exact host and a covering wildcard are not ambiguous, so I do not think this should be treated as a conflict.

Gateway API prescribes the opposite handling. `Listener.TLS`: "The GatewayClass MUST use the longest matching SNI out of all available certificates for any TLS handshake." And v1.3 added the `OverlappingTLSConfig` listener condition, whose own example is `foo.example.com` vs `*.example.com`: controllers MUST detect that overlap and set a condition on both listeners — the resources stay admitted and both keep working.

APISIX already behaves that way: a reversed exact SNI is a fully static path, so `match_route` finds it in `hash_path` before it ever walks the tree holding the wildcard prefixes. `app.example.com` gets its own certificate and `other.example.com` gets the wildcard one, every time.

The cost of calling it a conflict is concrete: "wildcard cert plus a dedicated cert for one subdomain" stops being expressible, and since `ValidateUpdate` re-runs the full detection over the whole object, existing Ingresses in that shape can no longer be edited at all after an upgrade — an unrelated annotation change gets rejected, and the only way out is deleting the other resource.

The case worth keeping is two objects carrying the *same* SNI: those land in one `hash_path[sni_rev]` array and `sort_route` ties on both priority and `path_org` length, so the winner is insertion order. Exact-key matching already covers it, which is why I would leave that part as it is and drop the overlap handling.

internal/webhook/v1/ssl/conflict_detector.go
```diff
@@ -478,7 +572,12 @@ func (d *ConflictDetector) mappingForHostWithCache(ctx context.Context, obj clie
 	}
 
 	for _, mapping := range mappings {
-		if mapping.Host == host {
+		if mapping.Host == "" {
+			continue
+		}
+		// Wildcard-aware: an exact host and a covering wildcard overlap even
+		// though their index keys differ ("app.example.com" vs "*.example.com").
+		if sslutil.HostsOverlap(mapping.Host, host) {
```
AlinsRan, line 585: First overlap wins here, so the outcome depends on mapping order: an exact host using the very same certificate gets reported as a conflict.

Repro — existing Ingress with `tls[0]={hosts:[*.example.com], secret:cert-a}`, `tls[1]={hosts:[app.example.com], secret:cert-b}`, then an ApisixTls for `app.example.com` with `cert-b`. The wildcard entry is hit first, `cert-a != cert-b`, conflict.

If overlap matching stays, match an exact `mapping.Host == host` first and only fall back to overlap.

internal/webhook/v1/ssl/conflict_detector.go
```diff
@@ -410,6 +471,39 @@ func (d *ConflictDetector) findExternalConflicts(ctx context.Context, obj client
 	return results, nil
 }
 
+// listAllTLSResources enumerates every TLS-bearing Gateway, Ingress and
+// ApisixTls. Used when the incoming host is a wildcard, whose covered exact
+// hosts can't be resolved through the exact-key host index.
+func (d *ConflictDetector) listAllTLSResources(ctx context.Context) ([]client.Object, error) {
```
AlinsRan, line 482: This lists every Gateway, Ingress and ApisixTls in the cluster and then runs `resolveGatewayProxy` (→ `FindMatchingIngressClass`) per candidate, on every admission whose host starts with `*.`. Wildcard hosts are the common case for TLS, so this is the common path, and it runs inside the admission timeout.

If overlap detection is kept, index for it instead: have `ssl_host.go` emit `ParentWildcard(host)` as an extra index key for exact hosts, and a `*.example.com` lookup resolves through the index. This function and the extra exact-side List then both go away.

internal/webhook/v1/ssl/conflict_detector.go
```diff
@@ -92,22 +98,23 @@ func (d *ConflictDetector) DetectConflicts(ctx context.Context, obj client.Objec
 	conflicts := make([]SSLConflict, 0)
 
 	// First, check for conflicts within the new resource itself.
-	seen := make(map[string]string, len(newMappings))
+	seen := make(map[string]HostCertMapping, len(newMappings))
```
AlinsRan, line 101: The self-conflict check still compares exact `mapping.Host`, so it misses what the external path now catches. A single Ingress with `tls[0]={hosts:[*.example.com], secret:a}` and `tls[1]={hosts:[app.example.com], secret:b}` passes. Whichever matching rule you settle on, both loops should use it.

internal/adc/translator/apisixtls.go
```diff
@@ -50,6 +50,14 @@ func (t *Translator) TranslateApisixTls(tctx *provider.TranslateContext, tls *ap
 		return nil, err
 	}
 
+	// APISIX serves the cert regardless of SAN, so this is advisory: warn when a
+	// declared host isn't covered by the cert SANs (clients may reject it).
+	if uncovered, sans := uncoveredSNIHosts(cert, tls.Spec.Hosts); len(uncovered) > 0 {
```
AlinsRan, line 55: Move this to the ApisixTls webhook as an `admission.Warnings`. APISIX never looks at cert SANs when selecting an SSL object, so this is purely advice for the user, and a log line in the translator is not something they will see — `ValidateCreate`/`ValidateUpdate` already return warnings that show up on `kubectl apply`. It also avoids re-parsing the certificate on every reconcile.

The PR description says translation is rejected when a host is not covered; the code only logs. Worth updating.

AlinsRan (changes requested):

AlinsRan (changes requested):
We are after the same thing: no host should quietly get served a certificate the operator did not intend. My concern is that admission rejection cannot deliver that, and that it costs existing users something real.

**Rejecting at admission does not make the data plane deterministic.** When two SSL objects carry the same SNI they land in one `hash_path[sni_rev]` array, and `sort_route` ties on both priority and `path_org` length, so the served certificate is decided by insertion order. Admission blocks one way in; it does not change what happens once a conflict exists.

**Conflicts arrive without an admission event.** A certificate rotated inside the Secret makes two previously-agreeing resources diverge. A GatewayProxy or IngressClass change moves a resource into a different group. The webhook can be disabled, and anything created before it was installed never passed through it. `ValidateCreate` / `ValidateUpdate` see none of these.

**The gate has a cost on existing clusters.** `ValidateUpdate` runs the full detection over the whole object, not just the changed fields. After an upgrade, an Ingress that has always coexisted with a wildcard resource can no longer be edited at all — an unrelated annotation change gets rejected — and the only way out is deleting the other resource. That breaks configuration that was never wrong.

**Gateway API defines this case and prescribes the opposite response.** `Listener.TLS`: "The GatewayClass MUST use the longest matching SNI out of all available certificates for any TLS handshake." v1.3 then added the `OverlappingTLSConfig` listener condition — "Controllers MUST detect the presence of overlapping hostnames... This condition MUST be set on all Listeners with overlapping TLS config" — and the spec's own example is `foo.example.com` vs `*.example.com`. Tie-breaking is specified too: oldest `creationTimestamp`, then first alphabetically by `namespace/name`. We do not set that condition anywhere today.

| | this PR | report + resolve |
|---|---|---|
| `*.example.com` and `app.example.com` | rejected at admission | admitted; longest SNI wins, as the spec requires |
| same SNI on two objects | rejected when admission can see it | admitted; oldest `creationTimestamp` wins |
| conflict from cert rotation / proxy change / pre-existing object | invisible | found on the next reconcile |
| how the operator learns about it | one error at apply time | condition on both objects, plus a webhook warning at apply time |
| what the data plane serves | unchanged, insertion order | one certificate per SNI, by a stated rule |

The last row is the one worth weighing: this PR adds a gate and leaves what the data plane serves exactly as it was.

Can we agree on the direction before iterating further on the implementation?

shreemaan-abhishek:
Agreed on the direction — thanks for the detailed writeup.

You're right on the two points that matter: admission rejection is incomplete (it can't see a conflict introduced by a certificate rotating inside its Secret, a GatewayProxy/IngressClass regrouping, a disabled webhook, or an object created before the webhook existed), and it's a backward-compatibility break (an unrelated edit to an object that already coexists with an overlapping one gets rejected). And overlapping SNIs are what Gateway API says to admit-and-report, not deny.

I've pushed a change so the detector now emits an admission **warning** instead of denying, on both `ValidateCreate` and `ValidateUpdate`. The detection itself — wildcard/exact overlap and the mTLS client-config dimension — stays, so the warning is accurate; the data plane behavior is unchanged, as you noted.

The fuller treatment you describe — detect on **reconcile**, set an `OverlappingTLSConfig`-style status condition on the affected objects, with the documented tie-break (longest SNI, then oldest `creationTimestamp`, then `namespace/name`) — is a larger change; I'd like to take it as a tracked follow-up rather than grow this PR. Does splitting it that way work for you?

One clarification for context: reject-at-admission predates this PR. This change extended what the detector notices (wildcard overlap + mTLS config) and now moves the response from deny to warn.

internal/webhook/v1/apisixtls_webhook.go
```diff
@@ -74,16 +74,17 @@ func (v *ApisixTlsCustomValidator) ValidateCreate(ctx context.Context, obj runti
...
-	conflicts := detector.DetectConflicts(ctx, tls)
-	if len(conflicts) > 0 {
-		return nil, fmt.Errorf("%s", sslvalidator.FormatConflicts(conflicts))
-	}
-
 	warnings := v.collectWarnings(ctx, tls)
+	// Overlapping SSL config is advisory: APISIX serves overlapping objects and
+	// resolves by SNI, so warn instead of denying.
+	skipADCValidation := v.initErr != nil || len(warnings) > 0
+	if conflicts := detector.DetectConflicts(ctx, tls); len(conflicts) > 0 {
```
nic-6443, line 81: [P1] Apply the warning behavior to `ValidateUpdate` too. This hunk changes only create; the update path still returns a hard error for the same conflict. An existing ApisixTls in an overlapping setup therefore still cannot be edited, which is one of the cases this PR says the warning conversion fixes. Please mirror this flow in `ValidateUpdate` and add an update regression test.

internal/webhook/v1/ssl/conflict_detector.go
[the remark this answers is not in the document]
nic-6443: The skip-regex hash collision is fixed, but this diagnostic is still unchanged: `FormatConflicts` says “different certificate” when only `ClientConfigHash` differs. Please make the warning generic or report which TLS component differs.

internal/webhook/v1/ssl/conflict_detector.go
nic-6443: Fixed by the length-prefixed regex encoding and its collision regression test. This concern is addressed.

PR #2811, opened 2026-07-19
fix: verify control-plane TLS certificate by default

shreemaan-abhishek:
## What this PR does

`GatewayProxy` `spec.provider.controlPlane.tlsVerify` is a `*bool` with no default. When omitted (as the shipped `examples/httpbin` manifests do, while pointing at `https://` Admin API endpoints), it decodes to `nil`, the translator leaves `cfg.TlsVerify` at the Go zero value `false`, and the executor inverts that into `tlsSkipVerify: true`. The result is that a default install skips certificate verification on the connection that carries the AdminKey and the full gateway config.

This makes verification the secure default:

- `+kubebuilder:default=true` on the CRD field, so an omitted `tlsVerify` becomes `true`.
- `cfg.TlsVerify = true` in the translator before honoring an explicit value, so behavior is correct even against an older CRD without the default. Only an explicit `tlsVerify: false` opts out.
- Regenerated CRD manifest and API reference docs; documented that setting it to `false` disables certificate verification.

## Testing

Added a translator unit test asserting the default: unset -> verify, explicit `false` -> skip, explicit `true` -> verify.

```
go test ./internal/adc/translator/ -run TestTranslateGatewayProxyToConfig_TlsVerifyDefault
```

### DIFF
--- a/api/v1alpha1/gatewayproxy_types.go
+++ b/api/v1alpha1/gatewayproxy_types.go
@@ -133,7 +133,10 @@ type ControlPlaneProvider struct {
 
 	Service *ProviderService `json:"service,omitempty"`
 	// TlsVerify specifies whether to verify the TLS certificate of the control plane.
+	// Defaults to true. Setting it to false disables certificate verification and
+	// exposes the AdminKey to man-in-the-middle attacks over https endpoints.
 	// +optional
+	// +kubebuilder:default=true
 	TlsVerify *bool `json:"tlsVerify,omitempty"`
 
 	// Auth specifies the authentication configuration.
--- a/config/crd/bases/apisix.apache.org_gatewayproxies.yaml
+++ b/config/crd/bases/apisix.apache.org_gatewayproxies.yaml
@@ -147,8 +147,11 @@ spec:
                         - name
                         type: object
                       tlsVerify:
-                        description: TlsVerify specifies whether to verify the TLS
-                          certificate of the control plane.
+                        default: true
+                        description: |-
+                          TlsVerify specifies whether to verify the TLS certificate of the control plane.
+                          Defaults to true. Setting it to false disables certificate verification and
+                          exposes the AdminKey to man-in-the-middle attacks over https endpoints.
                         type: boolean
                     required:
                     - auth
--- a/docs/en/latest/reference/api-reference.md
+++ b/docs/en/latest/reference/api-reference.md
@@ -312,7 +312,7 @@ ControlPlaneProvider defines configuration for control plane provider.
 | `mode` _string_ | Mode specifies the mode of control plane provider. Can be `apisix` or `apisix-standalone`. |
 | `endpoints` _string array_ | Endpoints specifies the list of control plane endpoints. |
 | `service` _[ProviderService](#providerservice)_ |  |
-| `tlsVerify` _boolean_ | TlsVerify specifies whether to verify the TLS certificate of the control plane. |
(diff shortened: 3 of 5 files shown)


api/v1alpha1/gatewayproxy_types.go
```diff
@@ -133,7 +133,10 @@ type ControlPlaneProvider struct {
 
 	Service *ProviderService `json:"service,omitempty"`
 	// TlsVerify specifies whether to verify the TLS certificate of the control plane.
+	// Defaults to true. Setting it to false disables certificate verification and
+	// exposes the AdminKey to man-in-the-middle attacks over https endpoints.
 	// +optional
+	// +kubebuilder:default=true
```
AlinsRan, line 139: `ControlPlaneProvider` has no way to supply a CA — only Endpoints/Service/TlsVerify/Auth — and the ADC executor exposes just `TlsSkipVerify` (internal/adc/client/executor.go:81,337). With a self-signed or private-CA control plane, the only escape from verify=true is `tlsVerify: false`.

The likely outcome is that everyone hits the error and pastes `tlsVerify: false` into their manifests, which leaves us worse off than today: the opt-out becomes boilerplate nobody thinks about, and changing the default is no longer available as a lever.

Suggest landing `controlPlane.caBundle` (inline PEM or a Secret ref) first, then flipping the default in a release with an upgrade note.

AlinsRan (commented):
Agree with verifying by default — the population that configures `https://` at all is the population whose control plane sits outside the cluster, which is exactly where verification matters. In-cluster setups use plain `http://` on 9180 anyway (test/e2e/scaffold/k8s.go:316-329), so they are unaffected.

Two things to sort out first, though.

**1. This breaks the examples shipped in this repo.**
`examples/httpbin/quickstart.yaml:94`, `httproute.yaml:36`, `ingress.yaml:28`, `tcproute.yaml:36` all point at `https://127.0.0.1:7443` with no `tlsVerify`. The PR cites them as evidence of the insecure default but doesn't touch them, so after this they go from "insecure but working" to "cannot connect", failing inside ADC sync where it's hard to diagnose. They need an explicit `tlsVerify: false` plus a comment that it's a self-signed dev cert.

**2.** See the inline comment — there is no way to supply a CA.

Also needs an upgrade note: CRD defaulting applies to existing objects on read from etcd, so every GatewayProxy that omits `tlsVerify` flips to verify the moment the CRD is upgraded.

api/v1alpha1/gatewayproxy_types.go
nic-6443: Agreed; this is already a concrete regression in the shipped manifests. `examples/httpbin/{quickstart,httproute,ingress,tcproute}.yaml` use the documented `https://127.0.0.1:7443` endpoint without `tlsVerify`, so they stop syncing after this default changes. Since no CA material can be passed to ADC, a private or self-signed CA has no secure migration path. Please add CA support plus an upgrade path and update those examples before flipping the default.

shreemaan-abhishek: Addressed in the latest push:

- Control-plane CA support has landed in #2826, the ADC-side support has landed in api7/adc#552, and the controller now pins ADC 0.29.0.
- All four self-signed local examples explicitly set tlsVerify: false and label it as a development-only certificate.
- The upgrade guide now explains that CRD defaulting affects existing GatewayProxy objects and documents the system-CA, caCert, and temporary development opt-out migration paths.
- The branch is merged with current master and the focused translator tests pass.

apache/apisix-helm-chart#994 remains the coordinated release-packaging follow-up; no controller release containing this new default exists yet.

AlinsRan (approved):

PR #2813, opened 2026-07-19
fix: fail loud when enable-csrf is set but csrf-key annotation is missing

shreemaan-abhishek:
## What this PR does

Fixes FINDING-031. When `enable-csrf: "true"` is set on an Ingress but `csrf-key` is missing or empty, the csrf plugin was silently dropped: the route was programmed **without** CSRF protection and the Ingress reconciled cleanly, so an endpoint the operator believed was protected ran unprotected with no error surfaced.

## Approach (changed after review)

Enforce the rule in the **Ingress validating admission webhook** (`ingress_webhook.go`), for both `ValidateCreate` and `ValidateUpdate`. `enable-csrf: true` with an empty/missing `csrf-key` is now a hard error, so `kubectl apply` is rejected loudly and synchronously, before the object is ever stored, for both new and updated Ingresses.

This replaces the earlier translation-failure approach, per review feedback from @AlinsRan and @nic-6443. That approach was reverted because:

1. **It did not cover the update path.** If an Ingress already served a route without CSRF and was then edited to add `enable-csrf` with no key, `Provider.Update` returned on the translation error *before* `UpdateConfig`, so the old unprotected route stayed live in the data plane (raised by @nic-6443).
2. **It over-reached.** An Ingress syncs `{Service, SSL}` together, so failing the whole translation on any annotation error meant a typo in a non-security annotation (e.g. `upstream-scheme`) took down the Ingress including its TLS (raised by @AlinsRan).
3. **Dropping a route does not stop traffic** anyway: requests fall through to any broader wildcard route, which may have no CSRF (raised by @AlinsRan).

The webhook can enforce severity properly (hard error vs warning) and blocks the bad intent at the source, which the translation layer structurally cannot.

## Known limitation

The Ingress webhook is deployed with `failurePolicy: Ignore`, so if the webhook is unavailable the invalid Ingress is admitted. This is a best-effort gate, not an unbreakable boundary; hardening that bypass is a separate concern (same class as the webhook-availability finding).

## Test

Added to `ingress_webhook_test.go`:
- `enable-csrf` without a key, and with an empty key, are rejected on **create**.
- the same invalid combination is rejected on **update** (covers the already-served case).
- `enable-csrf` with a key, and csrf-not-enabled, are allowed.

```
go test ./internal/webhook/... ./internal/adc/...
```

### DIFF
--- a/internal/adc/translator/annotations/plugins/csrf.go
+++ b/internal/adc/translator/annotations/plugins/csrf.go
@@ -16,6 +16,8 @@
 package plugins
 
 import (
+	"fmt"
+
 	adctypes "github.com/apache/apisix-ingress-controller/api/adc"
 	"github.com/apache/apisix-ingress-controller/internal/adc/translator/annotations"
 )
@@ -39,7 +41,11 @@ func (c *csrf) Handle(e annotations.Extractor) (any, error) {
 
 	key := e.GetStringAnnotation(annotations.AnnotationsCsrfKey)
 	if key == "" {
-		return nil, nil
+		// csrf requested without a key: the webhook rejects this, but log a
+		// breadcrumb for the cases it cannot cover (webhook off, pre-upgrade
+		// Ingress). Parse skips only this plugin, the rest of the route stands.
+		return nil, fmt.Errorf("annotation %q is enabled but %q is missing or empty",
+			annotations.AnnotationsEnableCsrf, annotations.AnnotationsCsrfKey)
 	}
 
 	return &adctypes.CSRFConfig{
--- a/internal/webhook/v1/ingress_webhook.go
+++ b/internal/webhook/v1/ingress_webhook.go
@@ -77,14 +77,14 @@ func (v *IngressCustomValidator) ValidateCreate(ctx context.Context, obj runtime
 		return nil, nil
 	}
 
+	if err := validateAnnotations(ingress); err != nil {
+		return nil, err
+	}
+
 	detector := sslvalidator.NewConflictDetector(v.Client)
 	conflicts := detector.DetectConflicts(ctx, ingress)
 	if len(conflicts) > 0 {
 		return nil, fmt.Errorf("%s", sslvalidator.FormatConflicts(conflicts))
-	}
-
-	if err := validateAnnotations(ingress); err != nil {
-		return nil, err
 	}
 
 	warnings := v.collectReferenceWarnings(ctx, ingress)
@@ -102,14 +102,14 @@ func (v *IngressCustomValidator) ValidateUpdate(ctx context.Context, oldObj, new
 		return nil, nil
 	}
 
+	if err := validateAnnotations(ingress); err != nil {
+		return nil, err
+	}
+
 	detector := sslvalidator.NewConflictDetector(v.Client)
 	conflicts := detector.DetectConflicts(ctx, ingress)
 	if len(conflicts) > 0 {
 		return nil, fmt.Errorf("%s", sslvalidator.FormatConflicts(conflicts))
-	}
-
-	if err := validateAnnotations(ingress); err != nil {
-		return nil, err
 	}
 
 	warnings := v.collectReferenceWarnings(ctx, ingress)
--- a/internal/webhook/v1/ingress_webhook.go
+++ b/internal/webhook/v1/ingress_webhook.go
@@ -28,6 +28,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
+	"github.com/apache/apisix-ingress-controller/internal/adc/translator/annotations"
 	"github.com/apache/apisix-ingress-controller/internal/controller"
 	"github.com/apache/apisix-ingress-controller/internal/webhook/v1/reference"
 	sslvalidator "github.com/apache/apisix-ingress-controller/internal/webhook/v1/ssl"
@@ -82,6 +83,10 @@ func (v *IngressCustomValidator) ValidateCreate(ctx context.Context, obj runtime
 		return nil, fmt.Errorf("%s", sslvalidator.FormatConflicts(conflicts))
 	}
 
+	if err := validateAnnotations(ingress); err != nil {
+		return nil, err
+	}
+
 	warnings := v.collectReferenceWarnings(ctx, ingress)
 	return warnings, nil
 }
@@ -103,8 +108,26 @@ func (v *IngressCustomValidator) ValidateUpdate(ctx context.Context, oldObj, new
 		return nil, fmt.Errorf("%s", sslvalidator.FormatConflicts(conflicts))
 	}
 
+	if err := validateAnnotations(ingress); err != nil {
+		return nil, err
+	}
+
 	warnings := v.collectReferenceWarnings(ctx, ingress)
 	return warnings, nil
+}
+
+// validateAnnotations rejects annotation combinations that would otherwise be
+// silently dropped, leaving the route without a requested security plugin.
+// enable-csrf with no csrf-key is refused here so kubectl apply fails loudly
+// instead of programming a route the operator believes is protected.
+func validateAnnotations(ingress *networkingv1.Ingress) error {
+	e := annotations.NewExtractor(ingress.Annotations)
+	if e.GetBoolAnnotation(annotations.AnnotationsEnableCsrf) &&
+		e.GetStringAnnotation(annotations.AnnotationsCsrfKey) == "" {
+		return fmt.Errorf("annotation %q is enabled but %q is missing or empty",
+			annotations.AnnotationsEnableCsrf, annotations.AnnotationsCsrfKey)
+	}
+	return nil
 }
 
 // ValidateDelete implements webhook.CustomValidator so a webhook will be registered for the type Ingress.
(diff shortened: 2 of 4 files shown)


internal/adc/translator/annotations/plugins/csrf.go
```diff
@@ -39,7 +41,10 @@ func (c *csrf) Handle(e annotations.Extractor) (any, error) {
 
 	key := e.GetStringAnnotation(annotations.AnnotationsCsrfKey)
 	if key == "" {
-		return nil, nil
+		// csrf requested but the key is missing: fail loud instead of
+		// silently dropping the plugin and programming the route unprotected.
+		return nil, fmt.Errorf("annotation %q is enabled but %q is missing or empty",
+			annotations.AnnotationsEnableCsrf, annotations.AnnotationsCsrfKey)
```
AlinsRan, line 48: Error is right, but `plugins.Parse` (plugins.go:62-65) does `log.Error; continue` and returns nil, so it never reaches `translateAnnotations`' `errors.Join`. Route still programmed without csrf — only a log line, no status/event. Failing loud needs `plugins.Parse` to propagate the error. See top-level comment.

AlinsRan (commented):
Right direction, but the error is swallowed one level up, so it does not fail loud.

`plugins.Parse` (plugins.go:62-65) logs the handler error and `continue`s, returning `(plugins, nil)`. So `translateAnnotations` (annotations.go:70) never receives it, and the route is still programmed without csrf — no status, no event, just one controller log line. The finding's "no error, event" still holds; only "no log" is fixed.

To actually fail loud, propagate the error through `plugins.Parse` instead of `continue`, so `translateAnnotations` surfaces it via `errors.Join` onto status (fail-closed on the route). That flips behavior for all annotation handlers — a deliberate call, state it explicitly.

`TestCSRFHandler` only tests the handler in isolation, so it passes while the route still builds unprotected. Test through `Parse` to cover the real behavior.

internal/adc/translator/annotations/plugins/csrf.go
```diff
@@ -39,7 +41,10 @@ func (c *csrf) Handle(e annotations.Extractor) (any, error) {
 
 	key := e.GetStringAnnotation(annotations.AnnotationsCsrfKey)
 	if key == "" {
-		return nil, nil
+		// csrf requested but the key is missing: fail loud instead of
+		// silently dropping the plugin and programming the route unprotected.
```
nic-6443: This still fails open at the route level. plugins.Parse logs handler errors and continues, then returns a nil error, so this error never reaches reconciliation and the route is still programmed without the csrf plugin. Please propagate the handler error from plugins.Parse and cover the parser/reconcile path so enable-csrf=true cannot reconcile successfully without a key.

AlinsRan (commented):
But I'd suggest doing this in the Ingress webhook instead.

`internal/webhook/v1/ingress_webhook.go` already validates Ingress, and already uses two severities — a hard error for SSL conflicts, `admission.Warnings` for missing references. This fits:

- `enable-csrf` set with an empty `csrf-key` → **error**, rejected at `kubectl apply`.
- `upstream-scheme` / `upstream-retry` / timeouts / rewrite regex → **warning**, printed in the `kubectl apply` output.

Two reasons not to fail the whole translation:

1. An Ingress syncs `{TypeService, TypeSSL}` (provider/apisix/provider.go:130), so a typo in `upstream-retry` also withdraws that Ingress's TLS certificate.
2. Dropping the route doesn't stop the traffic — it falls through to whatever other Ingress has a wildcard host, which may have no csrf at all.

ingress-nginx makes the same split: ordinary annotation errors are ignored, only the `auth-*` family is fail-closed, and even then the location is kept and serves `return 503;` instead of being removed.

internal/adc/translator/annotations/plugins/csrf.go
nic-6443: Thanks, the error now reaches reconciliation, but the update path still fails open. If an existing Ingress has already programmed a route without CSRF and is then updated with `enable-csrf: "true"` but no key, `Provider.Update` returns on this error before calling `UpdateConfig`; it does not delete or replace the previous resources, so the old unprotected route remains in the ADC cache and data plane. The translator-only tests cannot catch this. Please reject the invalid update in admission or program a fail-closed replacement, and cover an update from an already-served route.

shreemaan-abhishek:
@AlinsRan agreed, and I have moved this to the Ingress webhook. `enable-csrf: true` with an empty/missing `csrf-key` is now a hard error in `ValidateCreate` and `ValidateUpdate`, rejected at `kubectl apply`. I reverted the translation-failure change entirely.

Both your points held up when I traced them:

- **Blast radius.** Failing the whole translation on any annotation error coupled a non-security typo to the entire Ingress. Since an Ingress syncs `{Service, SSL}` together, a bad `upstream-scheme` would have taken TLS down with it. Enforcing at admission with error-vs-warning severity is the right split; the translation layer is all-or-nothing per Ingress and structurally can't do that.
- **Route drop != fail closed.** Correct that a dropped route falls through to a broader wildcard route. The webhook avoids this by blocking the bad intent before it is ever stored, so the route never changes state.

One thing I did not carry over yet: turning `upstream-scheme`/retry/timeout/regex errors into admission **warnings**. That is a good improvement but it is a separate concern from this CSRF finding, so I left it out to keep the PR focused. Happy to do it as a follow-up.

Note for anyone reading: the Ingress webhook is deployed with `failurePolicy: Ignore`, so this is a best-effort gate rather than a hard boundary. Hardening that bypass is out of scope here.

internal/adc/translator/annotations/plugins/csrf.go
shreemaan-abhishek: @nic-6443 you were right about the update path, thanks. I confirmed `Provider.Update` returns on the translation error before `UpdateConfig`, so the previously-programmed unprotected route survived in the ADC cache and data plane. Translator-only tests could not see it.

I have moved enforcement to the Ingress admission webhook (`ValidateCreate` + `ValidateUpdate`), so `enable-csrf: true` with no key is now rejected at apply time for updates as well as creates, and the already-served case is covered. The translation-failure change is reverted. Added a webhook test that exercises an update from a route with no csrf annotations to the invalid enable-csrf-without-key state.

internal/webhook/v1/ingress_webhook.go
```diff
@@ -82,6 +83,10 @@ func (v *IngressCustomValidator) ValidateCreate(ctx context.Context, obj runtime
 		return nil, fmt.Errorf("%s", sslvalidator.FormatConflicts(conflicts))
 	}
 
+	if err := validateAnnotations(ingress); err != nil {
```
nic-6443: [P1] This guard only runs when the admission webhook is enabled, but the shipped configuration defaults `webhook.enable` to `false`. In default deployments, `ValidateCreate` and `ValidateUpdate` never run and the translator still silently drops csrf, so both create and update remain unprotected. `failurePolicy: Ignore` is not the only bypass. Please enforce this invariant in a path that always runs, with fail-closed behavior for updates, or make admission mandatory for this feature.

AlinsRan: The shipped Helm chart defaults `webhook.enabled: true` (`charts/apisix-ingress-controller/values.yaml` in apisix-helm-chart). The `false` you are looking at is the Go default in `NewWebhookConfig()` / `config/samples/config.yaml`, which only applies when running the binary against a hand-written config.

Fail-closed on update was the previous revision of this PR and was reverted in the `csrf.go` thread: `Provider.Update` returns on the translation error *before* `UpdateConfig`, so the old unprotected route stays live anyway. It swaps one broken state for another rather than protecting anything.

The gap that is real is that the translator says nothing at all when the webhook is off. A log there is enough — commented on `csrf.go`.

internal/adc/translator/annotations/plugins/csrf.go
AlinsRan: Let's still return an error here when the key is missing.

Not to fail the route — `plugins.Parse` logs handler errors and skips only that plugin, so the rest of the Ingress is untouched. The webhook stays the enforcement point; this just means the cases it cannot cover (webhook disabled, or an Ingress created before the upgrade that never gets re-admitted) leave a log line instead of nothing at all.

webhook + log is enough here.

internal/webhook/v1/ingress_webhook.go
AlinsRan: Move this above `DetectConflicts`. It is a pure in-memory check, no reason to run it after several List calls and Secret reads.

internal/adc/translator/annotations/plugins/csrf.go
shreemaan-abhishek: Done in dc0fc9c. `csrf.Handle` now returns an error when the key is missing, so `plugins.Parse` logs a breadcrumb for the cases the webhook cannot cover (webhook off, or a pre-upgrade Ingress). Parse skips only the csrf plugin, so the rest of the route is untouched. The webhook stays the enforcement point. Kept it to webhook + log as suggested.

internal/webhook/v1/ingress_webhook.go
shreemaan-abhishek: Done in af013d93, moved above `DetectConflicts` in both `ValidateCreate` and `ValidateUpdate`.

AlinsRan (approved):

internal/webhook/v1/ingress_webhook.go
nic-6443: You are right: the shipped Helm chart defaults `webhook.enabled` to true; the `false` value is only the binary config default. The current head rejects normal create/update and logs the documented `failurePolicy: Ignore` bypass cases, so I consider this concern addressed.

nic-6443 (approved):

PR #2814, opened 2026-07-20
fix: reject invalid plugin config instead of applying an empty one

shreemaan-abhishek:
### Type of change:

- [x] Bugfix

When a plugin's `config` fails to unmarshal into an object (for example a
non-object value like `config: ["10.0.0.0/8"]`, which the CRD accepts
because `config` uses `x-kubernetes-preserve-unknown-fields`), the error
was logged and discarded, and the plugin was still applied with an empty
(or secret-only) config. So an `ip-restriction` plugin with a malformed
`whitelist` silently became `ip-restriction: {}` on the published route,
enforcing nothing while the resource reconciled green. The same path
serves `ApisixGlobalRule`, so a malformed global plugin config was
applied gateway-wide in mutated form.

This changes `buildPluginConfig` to return an error on a malformed
config and propagates it through its apiv2 and Ingress callers (route
plugins, stream-route plugins, referenced `ApisixPluginConfig`,
`ApisixGlobalRule`, Ingress plugin-config annotation, `ApisixConsumer`).
Translation now fails instead of publishing a fabricated empty object,
so the resource keeps its last good state.

On these paths the failure is observable. `ApisixRoute` and
`ApisixConsumer` go through `adcValidator`, which runs the real
translator inside the admission webhook, so an invalid config is
rejected at apply time. `ApisixGlobalRule` and `Ingress` have no such
validator, but both controllers call `Provider.Update` before writing
status, so the error lands in a status condition.

### Scope

The Gateway API paths (`httproute.go`, `grpcroute.go`) are deliberately
excluded. Gateway API requires that an `ExtensionRef` filter which
cannot be resolved must not be skipped, and that the affected requests
receive an HTTP error response instead. Failing translation does not
satisfy that: the route is never programmed, so a request either falls
through to another matching route and is served without the filter, or
404s. Those paths need the `fault-injection` treatment the translator
already applies to unresolvable `backendRefs`, scoped per rule. Tracked
in #2841, which depends on #2840 for the failure to be visible in
status.

The remaining log-and-skip sites (`consumer.go`, `gateway.go`,
`policies.go`) are also out of scope: each needs its own decision about
how the failure surfaces, since v1alpha1 `Consumer` has an
`adcValidator` case while `L4RoutePolicy` has no webhook at all.
Tracked in #2842.

### Pre-submission checklist:

### DIFF
--- a/internal/adc/translator/apisixconsumer.go
+++ b/internal/adc/translator/apisixconsumer.go
@@ -101,7 +101,10 @@ func (t *Translator) TranslateApisixConsumer(tctx *provider.TranslateContext, ac
 		if !plugin.Enable {
 			continue
 		}
-		config := t.buildPluginConfig(plugin, ac.Namespace, tctx.Secrets)
+		config, err := t.buildPluginConfig(plugin, ac.Namespace, tctx.Secrets)
+		if err != nil {
+			return nil, err
+		}
 		plugins[plugin.Name] = config
 	}
 
--- a/internal/adc/translator/httproute.go
+++ b/internal/adc/translator/httproute.go
@@ -83,13 +83,13 @@ func (t *Translator) fillPluginFromExtensionRef(plugins adctypes.Plugins, namesp
 		}
 		for _, plugin := range pluginconfig.Spec.Plugins {
 			pluginName := plugin.Name
-			pluginconfig := make(map[string]any)
+			config := make(map[string]any)
 			if len(plugin.Config.Raw) > 0 {
-				if err := json.Unmarshal(plugin.Config.Raw, &pluginconfig); err != nil {
+				if err := json.Unmarshal(plugin.Config.Raw, &config); err != nil {
 					return fmt.Errorf("failed to unmarshal config of plugin %s: %w", plugin.Name, err)
 				}
 			}
-			plugins[pluginName] = pluginconfig
+			plugins[pluginName] = config
 		}
 		t.Log.V(1).Info("fill plugin from extension ref", "plugins", plugins)
 	}
--- a/internal/adc/translator/httproute.go
+++ b/internal/adc/translator/httproute.go
@@ -45,7 +45,7 @@ func (t *Translator) fillPluginsFromHTTPRouteFilters(
 	filters []gatewayv1.HTTPRouteFilter,
 	matches []gatewayv1.HTTPRouteMatch,
 	tctx *provider.TranslateContext,
-) error {
+) {
 	for _, filter := range filters {
 		switch filter.Type {
 		case gatewayv1.HTTPRouteFilterRequestHeaderModifier:
@@ -59,19 +59,16 @@ func (t *Translator) fillPluginsFromHTTPRouteFilters(
 		case gatewayv1.HTTPRouteFilterResponseHeaderModifier:
 			t.fillPluginFromHTTPResponseHeaderFilter(plugins, filter.ResponseHeaderModifier)
 		case gatewayv1.HTTPRouteFilterExtensionRef:
-			if err := t.fillPluginFromExtensionRef(plugins, namespace, filter.ExtensionRef, tctx); err != nil {
-				return err
-			}
+			t.fillPluginFromExtensionRef(plugins, namespace, filter.ExtensionRef, tctx)
 		case gatewayv1.HTTPRouteFilterCORS:
 			t.fillPluginFromHTTPCORSFilter(plugins, filter.CORS)
 		}
 	}
-	return nil
 }
 
-func (t *Translator) fillPluginFromExtensionRef(plugins adctypes.Plugins, namespace string, extensionRef *gatewayv1.LocalObjectReference, tctx *provider.TranslateContext) error {
+func (t *Translator) fillPluginFromExtensionRef(plugins adctypes.Plugins, namespace string, extensionRef *gatewayv1.LocalObjectReference, tctx *provider.TranslateContext) {
 	if extensionRef == nil {
-		return nil
+		return
 	}
 	if extensionRef.Kind == internaltypes.KindPluginConfig {
 		pluginconfig := tctx.PluginConfigs[types.NamespacedName{
@@ -79,21 +76,21 @@ func (t *Translator) fillPluginFromExtensionRef(plugins adctypes.Plugins, namesp
 			Name:      string(extensionRef.Name),
 		}]
 		if pluginconfig == nil {
-			return nil
+			return
 		}
 		for _, plugin := range pluginconfig.Spec.Plugins {
 			pluginName := plugin.Name
-			config := make(map[string]any)
+			pluginconfig := make(map[string]any)
 			if len(plugin.Config.Raw) > 0 {
-				if err := json.Unmarshal(plugin.Config.Raw, &config); err != nil {
-					return fmt.Errorf("failed to unmarshal config of plugin %s: %w", plugin.Name, err)
+				if err := json.Unmarshal(plugin.Config.Raw, &pluginconfig); err != nil {
+					t.Log.Error(err, "plugin config unmarshal failed", "plugin", plugin.Name)
+					continue
 				}
 			}
-			plugins[pluginName] = config
+			plugins[pluginName] = pluginconfig
 		}
 		t.Log.V(1).Info("fill plugin from extension ref", "plugins", plugins)
 	}
-	return nil
 }
 
 func (t *Translator) fillPluginFromURLRewriteFilter(plugins adctypes.Plugins, urlRewrite *gatewayv1.HTTPURLRewriteFilter, matches []gatewayv1.HTTPRouteMatch) {
@@ -679,9 +676,7 @@ func (t *Translator) TranslateHTTPRoute(tctx *provider.TranslateContext, httpRou
 
 		enableWebsocket, _ := t.translateBackendsToUpstreams(tctx, rule, httpRoute, service)
 
-		if err := t.fillPluginsFromHTTPRouteFilters(service.Plugins, httpRoute.GetNamespace(), rule.Filters, rule.Matches, tctx); err != nil {
-			return nil, err
-		}
+		t.fillPluginsFromHTTPRouteFilters(service.Plugins, httpRoute.GetNamespace(), rule.Filters, rule.Matches, tctx)
 
 		matches := rule.Matches
 		if len(matches) == 0 {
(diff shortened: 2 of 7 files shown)


shreemaan-abhishek:
Checked this against a real API server (envtest, CRDs from `config/crd/bases`) rather than reasoning from the Go semantics alone, and `config: null` is not reachable.

The API server prunes the explicit null before it is persisted, so `plugin.Config.Raw` comes back empty rather than `[]byte("null")`:

```
ApisixRoute        ACCEPTED (null)                        -> stored raw="" len=0
ApisixPluginConfig ACCEPTED (null)                        -> stored raw="" len=0
ApisixRoute        ACCEPTED (["10.0.0.0/8"])              -> stored raw="[\"10.0.0.0/8\"]" len=14
ApisixRoute        ACCEPTED ("oops")                      -> stored raw="\"oops\"" len=6
ApisixRoute        ACCEPTED ({"whitelist":["10.0.0.0/8"]}) -> stored raw="{\"whitelist\":[\"10.0.0.0/8\"]}" len=28
```

With an empty `Raw` the `len(plugin.Config.Raw) > 0` guard is false, so the unmarshal is skipped entirely and `config` stays the empty non-nil map from `make`. There is no nil map, so no nil-map write in the `secretRef` merge and no `null` published for the plugin. The same holds on the `ApisixPluginConfig` path that `fillPluginFromExtensionRef` reads.

The non-object cases the PR does target (array, string, number) are persisted verbatim, which is what the new tests cover.

The variable shadowing in `fillPluginFromExtensionRef` was a fair point and is fixed.

internal/adc/translator/httproute.go
```diff
@@ -59,38 +59,41 @@ func (t *Translator) fillPluginsFromHTTPRouteFilters(
...
 		for _, plugin := range pluginconfig.Spec.Plugins {
 			pluginName := plugin.Name
-			pluginconfig := make(map[string]any)
+			config := make(map[string]any)
 			if len(plugin.Config.Raw) > 0 {
-				if err := json.Unmarshal(plugin.Config.Raw, &pluginconfig); err != nil {
-					t.Log.Error(err, "plugin config unmarshal failed", "plugin", plugin.Name)
-					continue
+				if err := json.Unmarshal(plugin.Config.Raw, &config); err != nil {
+					return fmt.Errorf("failed to unmarshal config of plugin %s: %w", plugin.Name, err)
```
AlinsRan: This diverges from Gateway API, and in a way the codebase already has an answer for.

The spec's rule for a custom (`ExtensionRef`) filter that cannot be applied is that the route keeps matching and the affected requests get an error response — it must not resolve to "the filter is not there":

> If a reference to a custom filter type cannot be resolved, the filter MUST NOT be skipped. Instead, requests that would have been processed by that filter MUST receive a HTTP error response.

[gateway-api v1.6.0 `httproute_types.go#L847-L849`](https://github.com/kubernetes-sigs/gateway-api/blob/v1.6.0/apis/v1/httproute_types.go#L847-L849) ([rendered](https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io%2fv1.HTTPRouteFilter)). Same shape for backendRefs, where the failure semantics are explicitly per-request and local: [`httproute_types.go#L271-L282`](https://github.com/kubernetes-sigs/gateway-api/blob/v1.6.0/apis/v1/httproute_types.go#L271-L282).

Returning an error here means the route is never programmed, which is not equivalent: if another route also matches, the request is served **without** the filter — the outcome the sentence above forbids — and if nothing else matches it is a 404, indistinguishable from "no such route". It also widens a per-filter failure into a whole-route one (see the call site), and one bad `PluginConfig` takes down every route referencing it.

The conformant pattern is already implemented ~40 lines below, for backendRefs: keep the route, attach a `fault-injection` 500 — [`httproute.go#L663-L681`](https://github.com/apache/apisix-ingress-controller/blob/dad45c50c15152fdc71e875155cad1f039060cdb/internal/adc/translator/httproute.go#L663-L681), whose error is deliberately discarded at the call site precisely because it has been turned into a data-plane 500. Suggest reusing that here instead of failing translation, plus `ResolvedRefs=False` on the route. Given that, it may be cleanest to drop `httproute.go` / `grpcroute.go` from this PR and do them separately — the apiv2 paths don't carry these constraints.

internal/adc/translator/httproute.go
```diff
@@ -676,7 +679,9 @@ func (t *Translator) TranslateHTTPRoute(tctx *provider.TranslateContext, httpRou
 
 		enableWebsocket, _ := t.translateBackendsToUpstreams(tctx, rule, httpRoute, service)
 
-		t.fillPluginsFromHTTPRouteFilters(service.Plugins, httpRoute.GetNamespace(), rule.Filters, rule.Matches, tctx)
+		if err := t.fillPluginsFromHTTPRouteFilters(service.Plugins, httpRoute.GetNamespace(), rule.Filters, rule.Matches, tctx); err != nil {
```
AlinsRan: Two consequences of aborting here worth being explicit about:

- It drops **every rule** of the HTTPRoute, not just the rule whose filter failed. Gateway API scopes filter/backendRef failures to the requests that would have hit them, never to the whole route.
- The failure is invisible. `httproute_controller.go` writes the status (`Accepted=True` / `ResolvedRefs=True`) *before* calling `Provider.Update`, and an error from `Update` only requeues — so the route reports accepted while nothing is programmed. There is no admission gate on this path either: `httproute_webhook.go` only does `collectWarnings` (no `adcValidator`), and v1alpha1 `PluginConfig` has no webhook at all. So unlike the apiv2 paths, the user gets no signal from either side.

Before this change the state was wrong but observable-by-behavior; after it, status and data plane disagree silently. If the fail-hard shape is kept for now, the status update needs to move after `Provider.Update` (or the error be folded into the condition).

internal/adc/translator/apisixconsumer.go
```diff
@@ -101,7 +101,10 @@ func (t *Translator) TranslateApisixConsumer(tctx *provider.TranslateContext, ac
 		if !plugin.Enable {
 			continue
 		}
-		config := t.buildPluginConfig(plugin, ac.Namespace, tctx.Secrets)
+		config, err := t.buildPluginConfig(plugin, ac.Namespace, tctx.Secrets)
```
AlinsRan, line 104: This makes apiv2 `ApisixConsumer` fail hard while its v1alpha1 counterpart keeps the old behavior — `consumer.go` logs and `continue`s for both the credential config and the consumer plugins, so a malformed credential is dropped and the consumer is published without it.

Still on the log-and-skip path after this PR:

- `consumer.go` — v1alpha1 Consumer credential config, and Consumer plugins
- `gateway.go` — GatewayProxy `plugins` and `pluginMetadata` (gateway-wide, the same blast radius the PR description cites for `ApisixGlobalRule`)
- `policies.go` — L4RoutePolicy plugins

The failure mode differs (they `continue`, so the plugin vanishes rather than becoming `{}`), but the user-visible outcome is the same class: the plugin does not run and everything reconciles green. Either sweep them in the same PR or say in the description why they are out of scope — otherwise the tree ends up with two deliberate, opposite conventions for the same situation.

AlinsRan (commented):
Fail-closed is the right instinct, and the apiv2 paths (ApisixRoute / ApisixPluginConfig / ApisixGlobalRule / ApisixConsumer / Ingress annotation) look good: there admission denies the change upfront (`apisixroute_webhook.go` wires `adcValidator.Validate`) and the reconciler reports the error (`defer r.updateStatus(&ar, err)`), so the failure is visible. Two things about the change as a whole:

**1. The Gateway API paths (`httproute.go` / `grpcroute.go`) need different handling, not the same fail-hard.** Details inline — the spec requires the route to keep matching and the affected requests to get an HTTP error response, and this repo already implements exactly that pattern for the analogous backendRef case. Suggest splitting those two files out of this PR.

**2. The root cause is the CRD schema, and it can be closed at the API server.** `Config apiextensionsv1.JSON` generates only `x-kubernetes-preserve-unknown-fields: true` with no `type`, which is why `config: ["10.0.0.0/8"]` is accepted in the first place (see the generated schema in `config/crd/bases/apisix.apache.org_apisixpluginconfigs.yaml`). Adding `+kubebuilder:validation:Type=object` to those `Config` fields emits `type: object` alongside the preserve-unknown-fields marker, so a non-object value is rejected by the API server with a proper field path — no webhook needed (the webhook is `failurePolicy=Ignore` and optional), and it covers the resources that have no webhook at all (v1alpha1 `PluginConfig` has none). The translator error is still worth keeping as defense in depth, since the CRDs in a cluster can be older than the controller. Worth a release note either way: it tightens the accepted input.

FYI the failing `e2e-test (apisix-standalone, networking.k8s.io)` is `TLSRoute Base Basic`, unrelated to this change (TLSRoute translation has no plugin config) and a known flake on apisix-standalone.

shreemaan-abhishek:
Thanks, all three land. Dropped `httproute.go` and `grpcroute.go` from this PR; it is now the apiv2 and Ingress paths only.

**Gateway API conformance.** Agreed, and returning an error is the worse of the two wrong answers: the route is never programmed, so a request either falls through to another matching route and is served without the filter (the outcome the spec forbids) or 404s. The `backendErr` → `fault-injection` 500 block is the right shape to reuse, scoped per rule with `ResolvedRefs=False`.

**Blast radius and visibility.** Confirmed both halves. `TranslateHTTPRoute` aborts on the first bad filter, so every rule of the route goes, and `Reconcile` persists the status before calling `Provider.Update`, so a translation error only requeues while the object still reports `Accepted=True`. `grpcroute_controller.go` has the same ordering. That one is worth fixing on its own regardless of this PR, since it makes any failure on those paths invisible.

For what it is worth, the paths that remain here do surface the failure: `ApisixRoute` and `ApisixConsumer` run the real translator through `adcValidator` in the admission webhook, and `ApisixGlobalRule` and `Ingress` call `Provider.Update` before writing status, so the error reaches a condition. That asymmetry is exactly why the split makes sense.

**Two conventions.** Fair, and I would rather not leave it implicit, so the description now states what is excluded and why. The remaining sites need individual decisions rather than one sweep: v1alpha1 `Consumer` has an `adcValidator` case and could fail closed like apiv2 `ApisixConsumer`, but `GatewayProxy` and `L4RoutePolicy` need their failure to be observable first, and `L4RoutePolicy` has no webhook at all.

One aside from the earlier `config: null` discussion: `mergeL4PolicyPlugins` normalizes a nil `cfg`, but that branch is unreachable. For a field typed `apiextensionsv1.JSON` with `x-kubernetes-preserve-unknown-fields` and no `nullable: true`, the API server prunes the explicit null, so `Config.Raw` arrives empty and the `len(Raw) > 0` guard skips the unmarshal. Verified with envtest against the generated CRDs on both `ApisixRoute` and `ApisixPluginConfig`. Worth removing whenever that file is touched.

AlinsRan (approved):

nic-6443 (approved):

PR #2818, opened 2026-07-21
fix: honor TCPRoute/UDPRoute sectionName and listener port for StreamRoute matching

### LINKED ISSUE #2802: bug: Gateway API TCPRoute ignores parentRefs.sectionName / listener port, causing all TCP listeners to match the same backend
### Current Behavior

The TCPRoute matches all configured TCP listeners.
Example:
9093 -> backend
9092 -> backend
although sectionName only references test1.

### Expected Behavior

A TCPRoute bound to:
parentRefs:
- name: apisix
  sectionName: test1
should only match the corresponding Gateway listener.
Example:
Gateway
listener test1 -> 9093

Gateway
listener test2 -> 9092

TCPRoute
parentRefs:
  sectionName: test1
Expected:
9093 -> backend
9092 -> connection refused / no route

### Error Logs

Experiment 1

Delete TCPRoute #2.

Result:

9092 still reaches backend

Experiment 2

Delete Gateway #2.

Result:

All TCP routes disappear.

Re-apply TCPRoute.

Result:

9092 -> backend
9093 -> backend

This indicates that reconciliation is working, but the generated StreamRoute is not restricted to the selected Gateway listener.

### Steps to Reproduce

Create Gateway #1

listener
9093

Create Gateway #2

listener
9092

Create only ONE TCPRoute

parentRefs:
  sectionName: test1

Result:

9093 -> backend
9092 -> backend

### Environment

**APISIX Version**
APISIX: 3.17.0
APISIX Ingress Controller: 2.1.0
Kubernetes: 1.31
Deployment mode: API-driven

AlinsRan:
### Type of change:

- [x] Bugfix

Fixes #2802

`TranslateTCPRoute` / `TranslateUDPRoute` emitted a single `StreamRoute` with **no match criteria** (the old `// TODO: ... server_port`). In APISIX stream mode a StreamRoute without `server_port` matches every connection on any stream listener, so all L4 routes across all Gateways collided onto one backend — `parentRefs.sectionName` and per-listener port routing were ignored.

This makes TCPRoute/UDPRoute participate in the existing `listener_port_match_mode` mechanism (added for HTTPRoute/GRPCRoute in #2703), so an L4 route's StreamRoute carries a `server_port` derived from the matched Gateway listener port(s):
- **Controllers** (`tcproute_controller.go`, `udproute_controller.go`): populate `tctx.Listeners` from the listeners `ParseRouteParentRefs` already matched for each parentRef (honoring `sectionName`, `port`, protocol, `allowedRoutes`) — mirroring the HTTPRoute/GRPCRoute path.
- **Translator** (`tcproute.go`, `udproute.go`): `buildL4StreamRoutes` emits one StreamRoute per matched listener port with `server_port` set and a port-qualified name/ID, gated by the same `shouldInjectServerPortVars`.

### Backward compatibility

`listener_port_match_mode` defaults to **`off`**, so by default L4 behavior is unchanged (a single portless StreamRoute) — no upgrade risk. Isolation is opt-in: set `listener_port_match_mode: auto` (or `explicit`) **and** ensure the Gateway listener port equals APISIX's physical `stream_proxy` port (e.g. 9100/9200), exactly as documented for HTTPRoute. This is the same, consistent contract across L4 and L7.

**Limitation:** `server_port` isolates routes across **distinct** listener ports only; two L4 routes on the *same* physical stream port with different backends still can't be disambiguated (bare TCP/UDP has no SNI/host).

### Pre-submission checklist:

- [x] Did you explain what problem does this PR solve?
- [x] Have you added corresponding test cases? (`l4route_serverport_test.go` — exercised in `auto` mode: explicit sectionName→port, multi-listener→distinct ServerPorts, dedup, single-no-explicit→portless, no-listener fallback; TCP + UDP)
- [ ] Have you modified the corresponding document? (n/a — reuses documented listener_port_match_mode)
- [x] Is this PR backward compatible? (default `off` → unchanged L4 behavior)

---
Rebased on latest master (Gateway API 1.6.0 #2804 + listener_port_match_mode default `off`). #2815/#2816/#2817 from this series are already merged.


internal/adc/translator/tcproute.go
```diff
@@ -42,6 +43,68 @@ func newDefaultUpstreamWithoutScheme() *adctypes.Upstream {
...
+//
+// Whether to inject server_port is gated by shouldInjectServerPortVars (the same
+// listener_port_match_mode used by HTTPRoute/GRPCRoute): the listener port is a
+// logical Gateway value that must equal APISIX's physical stream listen port for
+// the match to work, so injection is opt-in (explicit sectionName/port targeting,
+// or more than one listener port). When it is not injected we keep the previous
+// single portless StreamRoute, preserving backward compatibility.
+func (t *Translator) buildL4StreamRoutes(tctx *provider.TranslateContext, namespace, name string, ruleIndex int, typ, routeKind string, labels map[string]string) []*adctypes.StreamRoute {
+	var ports []int32
+	if portSet := listenerPortSet(tctx); t.shouldInjectServerPortVars(tctx.RouteParentRefs, portSet) {
```
nic-6443: [P2] Derive explicit targeting from matched parentRefs

`tctx.RouteParentRefs` is the unfiltered route spec, while `listenerPortSet` contains only listeners returned by `ParseRouteParentRefs`. If a route has an explicit parentRef that does not resolve or match plus a valid implicit parentRef with one listener, `hasExplicitListenerTarget` still returns true because of the invalid ref. In `auto` mode this injects the valid listener’s logical port instead of keeping the route portless, which can make the route unreachable when that port differs from APISIX’s physical stream port. Please derive the explicit-target signal from matched parent/listener contexts and cover the mixed invalid-explicit + valid-implicit case.

AlinsRan: Fixed in e081133a. `hasExplicitListenerTarget` now derives the explicit signal from the matched listeners (`tctx.Listeners`) instead of the raw parentRefs: a sectionName/port that does not resolve to a matched listener no longer counts as explicit, so a mix of an invalid explicit parentRef + a valid implicit single-listener parentRef stays portless in `auto` mode. `shouldInjectServerPortVars` now takes the matched listeners and the change is shared by HTTPRoute/GRPCRoute/TCPRoute/UDPRoute. Added the mixed invalid-explicit + valid-implicit case to `TestShouldInjectServerPortVars`.

nic-6443 (approved):

shreemaan-abhishek (approved):

PR #2819, opened 2026-07-21
fix: respond 500 for HTTPRoute rules with no backendRefs

AlinsRan:
### Problem

The Gateway API `HTTPRouteNoBackendRefs` conformance test requires a rule with omitted or empty `backendRefs` to explicitly respond with **500**. Currently such requests return **404** (verified failing in both `apisix` and `apisix-standalone` conformance jobs).

### Root cause

`translateBackendsToUpstreams` only injects the fault-injection 500 plugin when a backendRef *fails to resolve* (`backendErr != nil`). When a rule has **no** backendRefs, the loop never runs, `backendErr` stays `nil`, the service gets an empty upstream, and requests fall through to 404.

Confirmed by dumping the translator output: `NoBackendRefs` produces an empty upstream with **no** fault-injection, whereas the already-passing `NonExistent`/`UnknownKind` cases produce fault-injection 500 over the same empty upstream.

### Fix

Inject the fault-injection 500 also when the rule carries no backendRefs. A backend that resolves but has no ready endpoints is left untouched (keeps its own 503 handling). The resulting config is byte-identical to the already-passing `NonExistent` case, so removing the conformance skip is safe.

### Verification

- New unit test `TestTranslateHTTPRouteOmittedBackendRefs` (omitted + empty backendRefs → fault-injection 500).
- Removed `HTTPRouteNoBackendRefs` from the conformance skip list.
- Build / translator unit tests / lint clean.

Based on the Gateway API 1.6.0 branch (`feat/gateway-api-1.6.0`).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,7 +54,7 @@ GO_LDFLAGS ?= "-X=$(VERSYM)=$(VERSION) -X=$(GITSHASYM)=$(GITSHA) -X=$(BUILDOSSYM
 # gateway-api
 GATEAY_API_VERSION ?= v1.6.0
 ## https://github.com/kubernetes-sigs/gateway-api/blob/v1.6.0/pkg/features/httproute.go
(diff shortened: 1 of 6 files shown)


AlinsRan:
The omitted-backendRefs 500 fix (plus the redirect-filter regression guard it needed) landed in master via #2804. This PR is now redundant, closing.

PR #2820, opened 2026-07-21
fix: isolate same-port listeners by hostname instead of server_port

AlinsRan:
### Problem

The `HTTPRouteListenerHostnameMatching` and `GRPCRouteListenerHostnameMatching` conformance tests (a Gateway with 4 listeners sharing port 80, each with a distinct hostname, and HTTPRoutes bound to specific listeners via `sectionName`) returned **404 for every request**.

### Root cause

Because the routes target listeners via `sectionName`, `shouldInjectServerPortVars` treats them as explicit targets and `addServerPortVars` injects a `server_port == 80` var on every generated route (all listeners are on port 80).

`server_port` is the port APISIX actually accepts the connection on — its `node_listen` port (9080 in the test/e2e dataplane, fronted by a Service that maps external `:80 -> targetPort 9080`). It is **not** the Gateway's declared listener port. So `server_port == 80` never matches and all traffic 404s.

The existing `server_port` e2e coverage sidesteps this by deliberately using listener ports 9080/9081 that equal APISIX's `node_listen` (see the comment on `multiListenerGateway`). The conformance case uses port 80 and therefore exposed the gap.

Isolation here should be by **hostname** (each listener has a distinct hostname), which is already carried on `service.hosts` and proven to work for same-port host isolation by existing e2e tests. The `server_port` var adds no isolation for hostname-bearing listeners (the shared port cannot distinguish them) and actively breaks routing.

### Fix

Restrict `server_port` injection to the ports of listeners that have **no** hostname — the only listeners that rely on port-based isolation. Hostname-bearing listeners are left to host matching via `service.hosts`. Port-based routing for hostname-less listeners is unchanged.

Exact-vs-wildcard specificity (e.g. `foo.bar.com` must reach the `foo.bar.com` listener's backend, not the `*.bar.com` one) is already handled by the existing route-priority calculation.

### Changes

- Add `collectServerPortMatchPorts`, which excludes hostname-bearing listeners, and use it in HTTPRoute and GRPCRoute translation.
- Remove `HTTPRouteListenerHostnameMatching` and `GRPCRouteListenerHostnameMatching` from the conformance known-gaps skip list.
- Unit tests covering same-port listeners differing by hostname (no `server_port` emitted) and mixed hostname / hostname-less listeners (only hostname-less ports emitted).

### Verification

- `go build ./...`
- `go test ./internal/...`
- `golangci-lint cache clean && golangci-lint run` (0 issues)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,7 +54,7 @@ GO_LDFLAGS ?= "-X=$(VERSYM)=$(VERSION) -X=$(GITSHASYM)=$(GITSHA) -X=$(BUILDOSSYM
 # gateway-api
 GATEAY_API_VERSION ?= v1.6.0
 ## https://github.com/kubernetes-sigs/gateway-api/blob/v1.6.0/pkg/features/httproute.go
-SUPPORTED_EXTENDED_FEATURES = "HTTPRouteDestinationPortMatching,HTTPRouteMethodMatching,HTTPRoutePortRedirect,HTTPRouteRequestMirror,HTTPRouteSchemeRedirect,GatewayAddressEmpty,HTTPRouteResponseHeaderModification,GatewayPort8080,HTTPRouteHostRewrite,HTTPRouteQueryParamMatching,HTTPRoutePathRewrite,HTTPRouteBackendProtocolWebSocket"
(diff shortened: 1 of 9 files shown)


AlinsRan:
The same-port listener hostname isolation fix landed in master via #2804. This PR is now redundant, closing.

PR #2834, opened 2026-08-04
fix: delete data plane config when a route leaves a managed Gateway

### LINKED ISSUE #2833: bug: moving a route's parentRefs to a Gateway of another GatewayClass leaves orphaned config in the old data plane
## Current Behavior

When an HTTPRoute's `parentRefs` are repointed from a Gateway managed by controller A to a
Gateway managed by controller B (a different `GatewayClass.spec.controllerName`), controller B
picks the route up correctly — but **controller A never removes the configuration it had already
pushed**. Its data plane keeps serving the route indefinitely.

The chain, on `master` (`be4aaef`) and on the released `2.1.0`:

1. `httproute_controller.go:75` — the reconciler is registered with
   `For(&gatewayv1.HTTPRoute{})` and no controller-name predicate, so **every** controller
   instance watches **every** HTTPRoute. Repointing `parentRefs` bumps the generation, so
   `GenerationChangedPredicate` passes and controller A does reconcile the change.

2. `utils.go:357-359` — `ParseRouteParentRefs` skips parentRefs whose GatewayClass belongs to
   another controller:

   ```go
   if string(gatewayClass.Spec.ControllerName) != config.ControllerConfig.ControllerName {
       continue
   }
   ```

   For controller A the route's only remaining parentRef is now controller B's Gateway, so the
   returned slice is empty.

3. `httproute_controller.go:172-174` — the empty slice returns early with no cleanup:

   ```go
   if len(gateways) == 0 {
       return ctrl.Result{}, nil
   }
   ```

   `Provider.Delete` is only reachable from the `IgnoreNotFound` branch at :147, i.e. when the
   object is deleted from the cluster. There is no path for "this object used to be mine and no
   longer is".

4. `adc/client/client.go:290,303` — the periodic sync replays the in-memory store
   (`ConfigManager.List()` → `GetResources(name)`); it never re-derives state from Kubernetes.
   The orphaned service therefore gets re-pushed to the data plane every sync period, forever.

Restarting the controller pod clears it, because startup rebuilds the store from the informer's
initial list and the route no longer produces an entry for controller A.

The same early return exists in all five Gateway API route reconcilers:

```
httproute_controller.go:172
grpcroute_controller.go:190
tcproute_controller.go:283
udproute_controller.go:283
tlsroute_controller.go:283
```

### Why this is hard to notice

Every Kubernetes-level signal is correct and green while it happens:

- the HTTPRoute's `status.parents` lists only controller B, with `Accepted=True`
- controller B's Gateway shows the route in `status.listeners[].attachedRoutes`
- controller A's Gateway shows the correct (reduced) `attachedRoutes` count
- controller A logs no errors — its sync loop reports success every period
- GitOps tooling reports the manifests as fully synced

The stale configuration is visible **only** in the old data plane's own control API
(`:9090/v1/services`). In our case the orphaned routes carried public hostnames, so this failed
*open*: hosts that were supposed to have been taken off the public gateway stayed publicly
reachable for over 21 hours, with the label metadata and `modifiedIndex` still showing values
from weeks before the change.

### Prior art

PR #2543 ("fix: residual data issue when updating ingressClassName", merged 2025-09-08) fixed
exactly this failure mode for Ingress, and its description states the problem in general terms:

> the controller simply ignores the Ingress resource. However, this behavior is problematic: the
> corresponding configuration is not removed and remains in the data plane, which may result in
> incorrect routing or stale configurations.

The resulting code in `ingress_controller.go:159-166` calls `Provider.Delete` when the
IngressClass no longer matches. The Gateway API route reconcilers were never given the
equivalent.

## Expected Behavior

When a route no longer references any Gateway managed by this controller, the controller removes
the configuration it previously pushed for that route, so the data plane stops serving it —
matching the Ingress/IngressClass behaviour from #2543.

## Error Logs

None. That is a significant part of the problem: the controller logs no error at any log level,
and the ADC sync reports success on every tick while re-pushing the orphaned configuration.

## Steps to Reproduce

1. Install two APISIX ingress controller instances with distinct controller names, e.g.
   `apisix.apache.org/apisix-ingress-controller` and
   `apisix.apache.org/apisix-ingress-controller-internal`, each with its own GatewayClass and
   Gateway (`gateway-a` in namespace `apisix`, `gateway-b` in namespace `apisix-internal`).

2. Create an HTTPRoute with `parentRefs: [gateway-a]` and a hostname, and confirm it serves:

   ```console
   $ curl -H 'Host: demo.example.com' http://<gateway-a>/
   200
   ```

3. Change **only** the `parentRefs` to `gateway-b` (keep the same name/namespace) and apply.

4. Observe that controller B accepts the route and `gateway-b` serves it — and that `gateway-a`
   **still** serves it too:

   ```console
   $ curl -H 'Host: demo.example.com' http://<gateway-a>/
   200        # expected: 404
   ```

5. Confirm the orphan is in controller A's data plane, and that it is never removed:

   ```console
   $ kubectl port-forward -n apisix pod/<apisix-pod> 9090:9090
   $ curl -s localhost:9090/v1/services | jq -r '.[].value | "\(.labels["k8s/name"]) \(.hosts)"'
   demo ["demo.example.com"]        # still present, indefinitely
   ```

6. `kubectl rollout restart deploy/apisix-ingress-controller -n apisix` — the entry disappears,
   confirming the store is the only thing holding it.

The same reproduction applies to GRPCRoute, TCPRoute, UDPRoute and TLSRoute.

## Environment

- APISIX Ingress controller version: `2.1.0` (also reproduces on `master`, `be4aaef`)
- ADC version: `0.26.0`
- Mode: standalone / ADC (`config_provider: yaml`)
- Kubernetes cluster version: `v1.34.9`
- Gateway API version: `v1.3.0`

johannes-engler-mw:
### Type of change:

- [x] Bugfix

Fixes #2833.

When a Gateway API route's `parentRefs` are repointed at a Gateway belonging to a different
`GatewayClass.spec.controllerName`, the new controller picks the route up, but the previous
controller never removes the configuration it had already pushed. Its data plane keeps serving
the route indefinitely — until the controller process happens to restart.

`ParseRouteParentRefs` skips parentRefs whose GatewayClass belongs to another controller
(`utils.go:357-359`), so the reconciler ends up with an empty gateway list and returns early:

```go
if len(gateways) == 0 {
    return ctrl.Result{}, nil
}
```

`Provider.Delete` was only reachable from the `IgnoreNotFound` branch, i.e. when the object is
deleted from the cluster. There was no path for "this object used to be mine and no longer is".

Because the periodic sync replays the in-memory store (`ConfigManager.List()` →
`GetResources()`) rather than re-deriving state from Kubernetes, the orphaned service is
re-pushed to the data plane on every sync period. Restarting the controller is the only thing
that clears it, since startup rebuilds the store from scratch.

This PR calls `Provider.Delete` before that early return, in all five route reconcilers
(HTTPRoute, GRPCRoute, TCPRoute, UDPRoute, TLSRoute).

This is the same fix #2543 applied to Ingress for the equivalent `ingressClassName` case, whose
description states the general problem:

> the controller simply ignores the Ingress resource. However, this behavior is problematic: the
> corresponding configuration is not removed and remains in the data plane, which may result in
> incorrect routing or stale configurations.

`ingress_controller.go:159-166` has carried that cleanup since; the Gateway API reconcilers never
received the equivalent.

#### Notes for reviewers

- **`TypeMeta` is set before the call.** `Provider.Delete` derives both the store key
  (`utils.NamespacedNameKind`) and the resource labels (`label.GenLabel`) from
  `obj.GetObjectKind()`, which is empty on objects read through the client. The existing
  `IgnoreNotFound` branches already set it for the same reason.

- **No-op when there is nothing to remove.** `DeleteConfig` → `applyStoreChanges(args, true)`
  looks the key up in `ConfigManager` first; for a route this controller never translated, the
  delta is empty and neither the store nor ADC is touched. Routes belonging to other controllers
  therefore cost one map lookup.

- **One observation left alone deliberately.** `apisixProvider.Delete` calls `syncNotify()`
  unconditionally, so a reconcile of a route owned by another controller now nudges the sync
  loop even when the delta was empty. `syncNotify` is a non-blocking send to a buffered channel,
  so these coalesce, but making it conditional on a non-empty delta would mean changing the
  `Provider.Delete` signature to return the delta. That felt out of scope here — happy to follow
  up separately if you would prefer it in this PR.

### Pre-submission checklist:

The test is in `test/e2e/gatewayapi/httproute.go`, in the existing "HTTPRoute with Multiple
Gateway" context: it attaches an HTTPRoute to the additional gateway, then repoints `parentRefs`
at a Gateway whose GatewayClass names `apisix.apache.org/not-exist`. The route object itself is
never deleted, so it exercises the empty-gateway-list path rather than the deletion path.

Verified against a kind cluster by rebuilding the controller image from each revision and
re-running the same test binary:

```console
# controller built from master, test present
$ make e2e-test TEST_FOCUS="stop being served after moving"
  [FAIL] ... HTTPRoute should stop being served after moving to another controller's Gateway
  test/e2e/gatewayapi/httproute.go:461
  FAIL! -- 0 Passed | 1 Failed | 1 Pending | 232 Skipped

# controller built with this fix, same test
$ make e2e-test TEST_FOCUS="stop being served after moving"
  SUCCESS! -- 1 Passed | 0 Failed | 1 Pending | 232 Skipped
```

Line 461 is the closing assertion: every earlier step passes in both runs, including the one
that confirms the route is served through the additional gateway before the move. Without the
fix the gateway keeps answering `200` for the full 30s retry window after `parentRefs` move away.

### DIFF
--- a/internal/adc/client/client.go
+++ b/internal/adc/client/client.go
@@ -240,9 +240,10 @@ func (c *Client) Delete(ctx context.Context, args Task) error {
 	return c.applySync(ctx, args, delta)
 }
 
-func (c *Client) DeleteConfig(ctx context.Context, args Task) error {
-	_, err := c.applyStoreChanges(args, true)
-	return err
+// DeleteConfig removes the stored configuration for args.Key and reports what
+// it removed, so callers can skip a data plane sync when the key held nothing.
+func (c *Client) DeleteConfig(ctx context.Context, args Task) (StoreDelta, error) {
+	return c.applyStoreChanges(args, true)
 }
 
 func (c *Client) Validate(ctx context.Context, task Task) error {
--- a/internal/controller/apisixconsumer_controller_test.go
+++ b/internal/controller/apisixconsumer_controller_test.go
@@ -41,8 +41,10 @@ import (
 
 const testConsumerNamespace = "default"
 
-// recordingProvider records the objects passed to Delete and can be told to fail.
+// recordingProvider records the objects passed to Update and Delete, and can be
+// told to fail a delete. Shared by the reconciler tests in this package.
 type recordingProvider struct {
+	updated   int
 	deleted   []types.NamespacedName
 	deleteErr error
 }
@@ -50,6 +52,7 @@ type recordingProvider struct {
 func (p *recordingProvider) Register(string, *http.ServeMux) {}
 
 func (p *recordingProvider) Update(context.Context, *provider.TranslateContext, client.Object) error {
+	p.updated++
 	return nil
 }
 
--- a/internal/controller/gateway_controller_publishservice_test.go
+++ b/internal/controller/gateway_controller_publishservice_test.go
@@ -19,7 +19,6 @@ package controller
 
 import (
 	"context"
-	"net/http"
 	"testing"
 
 	"github.com/go-logr/logr"
@@ -41,23 +40,7 @@ import (
 	"github.com/apache/apisix-ingress-controller/api/v1alpha1"
 	"github.com/apache/apisix-ingress-controller/internal/controller/config"
 	"github.com/apache/apisix-ingress-controller/internal/controller/status"
-	"github.com/apache/apisix-ingress-controller/internal/provider"
 )
-
-// recordingProvider counts data plane pushes so tests can assert whether a
-// reconcile reached Provider.Update.
-type recordingProvider struct {
-	updated int
-}
-
-func (p *recordingProvider) Update(context.Context, *provider.TranslateContext, client.Object) error {
-	p.updated++
-	return nil
-}
-func (p *recordingProvider) Delete(context.Context, client.Object) error { return nil }
-func (p *recordingProvider) Start(context.Context) error                 { return nil }
-func (p *recordingProvider) NeedLeaderElection() bool                    { return false }
-func (p *recordingProvider) Register(string, *http.ServeMux)             {}
 
 type recordingUpdater struct {
 	updates []status.Update
--- a/internal/controller/grpcroute_controller.go
+++ b/internal/controller/grpcroute_controller.go
@@ -187,6 +187,17 @@ func (r *GRPCRouteReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	}
 
 	if len(gateways) == 0 {
+		// See the HTTPRoute reconciler: a route that no longer references a
+		// Gateway managed by this controller must have its previously pushed
+		// configuration removed, or the data plane keeps serving it.
+		gr.TypeMeta = metav1.TypeMeta{
+			Kind:       KindGRPCRoute,
+			APIVersion: gatewayv1.GroupVersion.String(),
+		}
+		if err := r.Provider.Delete(ctx, gr); err != nil {
+			r.Log.Error(err, "failed to delete grpcroute", "grpcroute", gr)
+			return ctrl.Result{}, err
+		}
 		return ctrl.Result{}, nil
 	}
 
--- a/internal/controller/grpcroute_controller.go
+++ b/internal/controller/grpcroute_controller.go
@@ -181,12 +181,17 @@ func (r *GRPCRouteReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		msg:    "Route is accepted",
 	}
 
-	gateways, err := ParseRouteParentRefs(ctx, r.Client, r.Log, gr, gr.Spec.ParentRefs)
+	gateways, unresolvedParents, err := ParseRouteParentRefs(ctx, r.Client, r.Log, gr, gr.Spec.ParentRefs)
 	if err != nil {
 		return ctrl.Result{}, err
 	}
 
 	if len(gateways) == 0 {
+		if unresolvedParents {
+			// See the HTTPRoute reconciler: an unresolvable parentRef leaves ownership
+			// unknown rather than disproven, so the data plane must be left alone.
+			return ctrl.Result{}, nil
+		}
 		// See the HTTPRoute reconciler: a route that no longer references a
 		// Gateway managed by this controller must have its previously pushed
 		// configuration removed, or the data plane keeps serving it.
--- a/internal/controller/httproute_controller.go
+++ b/internal/controller/httproute_controller.go
@@ -163,12 +163,18 @@ func (r *HTTPRouteReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		msg:    "Route is accepted",
 	}
 
-	gateways, err := ParseRouteParentRefs(ctx, r.Client, r.Log, hr, hr.Spec.ParentRefs)
+	gateways, unresolvedParents, err := ParseRouteParentRefs(ctx, r.Client, r.Log, hr, hr.Spec.ParentRefs)
 	if err != nil {
 		return ctrl.Result{}, err
 	}
 
 	if len(gateways) == 0 {
+		if unresolvedParents {
+			// A missing Gateway or GatewayClass leaves ownership unknown rather than
+			// disproven. GatewayClass is cluster-scoped, so while one is absent every
+			// route under it resolves empty and deleting would drain the data plane.
+			return ctrl.Result{}, nil
+		}
 		// The route does not reference any Gateway managed by this controller.
 		// It may have referenced one before, e.g. when its parentRefs are
 		// repointed at a Gateway belonging to another GatewayClass, so the
--- a/internal/controller/httproute_controller.go
+++ b/internal/controller/httproute_controller.go
@@ -169,6 +169,21 @@ func (r *HTTPRouteReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	}
 
 	if len(gateways) == 0 {
+		// The route does not reference any Gateway managed by this controller.
+		// It may have referenced one before, e.g. when its parentRefs are
+		// repointed at a Gateway belonging to another GatewayClass, so the
+		// configuration a previous reconcile pushed has to be removed. Without
+		// this the data plane keeps serving the route indefinitely.
+		// Provider.Delete derives the resource labels from the object Kind, which
+		// is empty on objects read through the client.
+		hr.TypeMeta = metav1.TypeMeta{
+			Kind:       KindHTTPRoute,
+			APIVersion: gatewayv1.GroupVersion.String(),
+		}
+		if err := r.Provider.Delete(ctx, hr); err != nil {
+			r.Log.Error(err, "failed to delete httproute", "httproute", hr)
+			return ctrl.Result{}, err
+		}
 		return ctrl.Result{}, nil
 	}
 
--- a/internal/controller/httproute_controller_test.go
+++ b/internal/controller/httproute_controller_test.go
@@ -1,0 +1,111 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package controller
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/go-logr/logr"
+	"github.com/stretchr/testify/assert"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	k8stypes "k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	gatewayv1 "sigs.k8s.io/gateway-api/apis/v1"
+
+	"github.com/apache/apisix-ingress-controller/internal/manager/readiness"
+)
+
+type noopReadier struct {
+	readiness.ReadinessManager
+}
+
+func (noopReadier) Done(client.Object, k8stypes.NamespacedName) {}
+
+// TestHTTPRouteReconcile_EmptyGateways covers the two ways a route resolves to no
+// Gateway of this controller. Only a resolved parent naming another controller
+// proves the route is not ours; an unresolvable parent must leave the data plane
+// alone, since a missing cluster-scoped GatewayClass empties the list for every
+// route under it at once.
+func TestHTTPRouteReconcile_EmptyGateways(t *testing.T) {
+	scheme := parentRefTestScheme(t)
+
+	route := &gatewayv1.HTTPRoute{
+		ObjectMeta: metav1.ObjectMeta{Namespace: "default", Name: "r"},
+		Spec: gatewayv1.HTTPRouteSpec{
+			CommonRouteSpec: gatewayv1.CommonRouteSpec{
+				ParentRefs: []gatewayv1.ParentReference{{Name: "gw"}},
+			},
+		},
+	}
+	gw := func(class string) *gatewayv1.Gateway {
+		return &gatewayv1.Gateway{
+			ObjectMeta: metav1.ObjectMeta{Namespace: "default", Name: "gw"},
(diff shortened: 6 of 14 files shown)


johannes-engler-mw:
A review pass raised whether the early return should also clear this controller's own
`status.parents` entry, per [RouteParentStatus](https://gateway-api.sigs.k8s.io/reference/api-spec/1.4/spec/#gateway.networking.k8s.io/v1.RouteParentStatus).
Worth recording what I found, since the answer is not obvious from the diff.

**In the migration this PR targets, the stale entry does not survive.** The status write is a
wholesale replacement rather than a merge:

```go
// TODO: diff the old and new status
hr.Status.Parents = make([]gatewayv1.RouteParentStatus, 0, len(gateways))
...
hCopy.Status = hr.Status
```

So when the *new* controller reconciles the moved route, it replaces the whole `parents` list
and the previous controller's entry goes with it. Confirmed on a live two-controller cluster
after exactly this migration — all four moved routes carry a single entry, the new owner's:

```console
$ kubectl get httproute -A -o json | jq -r '.items[] | "\(.metadata.name)  parents=\(.status.parents|length)  [\(.status.parents[]?.controllerName)]"'
gdpr-service            parents=1  [apisix.apache.org/apisix-ingress-controller-internal]
pii-data-service        parents=1  [apisix.apache.org/apisix-ingress-controller-internal]
rbac                    parents=1  [apisix.apache.org/apisix-ingress-controller-internal]
structure-data-service  parents=1  [apisix.apache.org/apisix-ingress-controller-internal]
```

**A stale entry does persist in one narrower case:** moving a route to a class no live
controller serves — which is what the e2e test does with `apisix.apache.org/not-exist`, since
that is the cheapest way to take a route out of scope in a single-controller test environment.
Nobody rewrites the status there, so the old entry remains. That behaviour is identical on
`master`; this PR does not introduce or worsen it, and the data-plane cleanup is independent
of it.

**The underlying issue is the wholesale replacement itself**, which the `TODO` above already
flags. It is arguably a larger spec deviation than a leftover entry: controller A overwrites
controller B's `RouteParentStatus`, so a route legitimately parented to Gateways owned by two
different controllers cannot hold a stable status at all. Fixing that means giving the status
updater per-controller merge semantics across all five route kinds — a change to shared status
handling that I would rather not fold into a data-plane cleanup.

Happy to open a separate issue for the merge semantics if that is useful, or to fold it in here
if maintainers prefer them together.

AlinsRan:
### `len(gateways) == 0` conflates "not mine" with "can't tell yet"

`ParseRouteParentRefs` reaches an empty list through three different `continue` paths:

- `utils.go:359` — controllerName mismatch. This is the case the PR targets; deleting is correct.
- `utils.go:343` — Gateway `NotFound`
- `utils.go:353` — GatewayClass `NotFound`

The last two mean "the parent could not be resolved", not "this route belongs to someone else". They currently get the same treatment.

The concrete failure is a GatewayClass that briefly disappears — GitOps prune-then-apply, a CRD upgrade, an accidental delete. GatewayClass is cluster-scoped, so while it is gone *every* route under *every* Gateway of that class resolves to an empty gateway list. On master those routes go stale but keep serving; with this PR their data plane config is actively deleted. That turns a stale-config window into an outage.

### That window is easy to hit, because routes are re-reconciled frequently

I assumed `WithEventFilter(predicate.GenerationChangedPredicate{})` would filter out EndpointSlice churn. It does not — EndpointSlice maintains `metadata.generation`:

```go
// k8s.io/kubernetes pkg/registry/discovery/endpointslice/strategy.go
endpointSlice.Generation = 1                      // :63
ogNewMeta.Generation = ogOldMeta.Generation + 1   // :82
```

`listHTTPRoutesByServiceRef` indexes purely on backend service refs with no ownership filter, so every pod scale / rollout / readiness flip enqueues *all* routes referencing that service, including ones owned by other controllers. Each of those now reaches `Provider.Delete`, and `apisixProvider.Delete` calls `defer d.syncNotify()` unconditionally — which triggers `Client.Sync`, a full push of the whole store to every data plane.

### Suggestion: filter at the predicate layer, the way Ingress already does

`MatchesIngressClassPredicate` (`utils.go:1889`) solves exactly this:

```go
predicateFuncs.UpdateFunc = func(e event.UpdateEvent) bool {
	return MatchesIngressClass(c, log, e.ObjectOld) || MatchesIngressClass(c, log, e.ObjectNew)
}
```

A route equivalent gives three properties at once:

| situation | outcome |
|---|---|
| route owned by another controller | old ✗ / new ✗ → never enqueued, zero cost |
| GatewayClass temporarily unresolvable | ownership check fails → not enqueued → data plane untouched |
| route just moved off my Gateway | old ✓ / new ✗ → admitted **exactly once** → this PR's delete branch runs |

The predicate does not replace the delete branch added here — the single admitted reconcile is what makes the cleanup happen. They are complementary.

One note: unlike `MatchesIngressClass`, which reads a single field, the route ownership check is a multi-hop walk (parentRefs → Gateway → GatewayClass → controllerName). Worth reusing that walk from `ParseRouteParentRefs` rather than writing a second one, so the two cannot drift.

One known blind spot, probably acceptable: if a GatewayClass's `controllerName` is edited in place, neither the route nor its Gateway changes and the route reconcilers do not watch GatewayClass, so nothing enqueues those routes. Changing `gatewayClassName` on the Gateway is covered by `listHTTPRoutesForGateway`.

johannes-engler-mw:
Thanks for the careful read. The first point lands and I want to fix it. On the second I'd like to
push back a little, because I don't think a `For()` predicate actually gives us the first two rows of
that table.

### The conflation is real

Agreed on `len(gateways) == 0`. Only the `controllerName` mismatch at `utils.go:369` means "not mine";
the Gateway `NotFound` at `:353` and the GatewayClass `NotFound` at `:363` mean "could not resolve".
GatewayClass being cluster-scoped is what makes it bite — while it is gone, *every* route under *every*
Gateway of that class resolves empty at once, and this PR turns that from stale-but-serving into
deleted. That is a regression I don't want to ship.

Worth noting the same conflation already exists on the Ingress side: `FindMatchingIngressClass`
returns a bare error for a missing IngressClass and for one belonging to another controller alike
(`utils.go:1747` vs `:1755`), and `ingress_controller.go:159` deletes on either. So #2543 shipped this
hazard too — it isn't specific to the Gateway API path, which is part of why I'd rather fix the
decision than replicate the predicate.

### Why I don't think the predicate can be the guard

`MatchesIngressClassPredicate` is attached to `For()` only (`ingress_controller.go:75`), so it filters
events on the object itself. Every other enqueue path in these controllers goes through
`handler.EnqueueRequestsFromMapFunc`, which never consults it:

- `listHTTPRoutesByServiceRef` (`httproute_controller.go:284`) — the EndpointSlice churn you describe
- `listHTTPRoutesForGateway` (`:388`) — indexes on `indexer.ParentRefs`, no ownership filter
- `listHTTPRoutesForGatewayProxy` (`:641`), `listHTTPRoutesByExtensionRef` (`:312`), the policy watches

Walking the table again with a route-equivalent predicate on `For()`:

| situation | actual outcome |
|---|---|
| route owned by another controller | filtered for route events only. Any EndpointSlice / Gateway / GatewayProxy event still enqueues it, still reaches `Provider.Delete`, still hits the unconditional `defer d.syncNotify()` at `provider/apisix/provider.go:232` |
| GatewayClass temporarily unresolvable | same — one endpoint readiness flip on any referenced backend enqueues the route and the delete branch runs |
| route just moved off my Gateway | admitted exactly once — this row holds |

The third row works, and it's a genuinely nice property. But the first two are the ones being asked to
prevent an outage and remove a cost, and in both the traffic arrives through a map func, which is
below the predicate. The Ingress controller has the same gap today via `listIngressesByService`.

### What I'd propose instead

Let the reconciler tell the cases apart, by having `ParseRouteParentRefs` report whether any parentRef
was left unresolved alongside the gateway list:

```go
if len(gateways) == 0 {
    if unresolved {
        // ownership unknown, not disproven — leave the data plane alone
        return ctrl.Result{}, nil
    }
    // every parentRef resolved, none named this controller
    ... Provider.Delete ...
}
```

One extra signal out of the walk that already runs, threaded through the five call sites. It holds
regardless of which watch enqueued the reconcile, which is the property the predicate can't offer, and
it satisfies your "reuse the walk rather than writing a second one" point by construction — there is
no second walk.

One consequence worth stating: on recovery, nothing re-enqueues those routes if only the GatewayClass
was deleted and recreated, since the route reconcilers don't watch GatewayClass — the same blind spot
you flagged for in-place `controllerName` edits. They'd stay stale until the next event on the route or
its Gateway. That matches master's behaviour, so it's not a regression, but a GatewayClass watch would
close both and I'm happy to open a follow-up for it.

### On the predicate itself

I'm happy to add it regardless — it removes real work for direct route events and moves these
controllers toward the Ingress shape. I'd just want it on top of the guard rather than instead of it.
And if the actual goal is to stop the EndpointSlice-driven churn, the filter has to go inside the map
funcs, which is a bigger change than a data-plane cleanup PR should carry.

So: guard only, guard + `For()` predicate, or all three? Happy to do any of them here — just let me
know which you'd rather review in one PR.

johannes-engler-mw:
@AlinsRan Please check my question and provide feedback, to this can be merged.

johannes-engler-mw:
@AlinsRan any update on this?

AlinsRan:
> the filter has to go inside the map funcs

You're right, I agree — the `For()` predicate can't cover the map-func enqueues.

Regarding the `syncNotify` point mentioned earlier: it might be worth fixing here rather than in a separate follow-up. It's unconditional (`provider.go:231`) and the consumer loop has no debounce (`provider.go:277`), so every call turns into a full `Client.Sync`. With this PR, EndpointSlice churn on another implementation's backends reaches the new branch through `listHTTPRoutesByServiceRef`, and each one of those resyncs our data planes.

It looks like a light fix if we want to address it now: `Provider.Delete` can keep its signature, because `Client.DeleteConfig` has a single production caller (`provider.go:232`) and already computes the delta before discarding it (`client.go:243`). We could have it return `(StoreDelta, error)` and gate `syncNotify` on `len(delta.Deleted) > 0`.

johannes-engler-mw:
Both done, pushed.

**Guard** (`efb5269b`): `ParseRouteParentRefs` now reports whether any parentRef was left
unresolved, and the five reconcilers skip the delete in that case. Only the `controllerName`
mismatch reaches `Provider.Delete`; Gateway/GatewayClass `NotFound` leaves the data plane alone.

**syncNotify** (`7580d05d`): `Client.DeleteConfig` returns the `StoreDelta` it already computed
and `apisixProvider.Delete` gates `syncNotify` on `len(delta.Deleted) > 0`. Single production
caller, so no other call sites moved.

One thing worth recording from writing the e2e: **the GatewayClass case is narrower than we both
assumed.** `gatewayclass_controller.go:89` holds `apisix.apache.org/gc-protection` for as long as
any Gateway references the class, requeueing every 5s with a `DeletionBlocked` event. So a prune
or an accidental `kubectl delete` only marks it `Terminating` — it stays readable and the lookup
still resolves. Reaching the empty list needs CRD removal, a restore, or the finalizer being
dropped. `Gateway` `NotFound` has no such protection, and neither does an informer cache miss,
so the guard still earns its place — the outage window is just smaller than "any GitOps prune".

Tests: unit coverage for both branches of the walk and for the reconciler decision, plus an e2e
that drops the finalizer to produce the real state (class gone, Gateway alive). Verified on kind:
passes with the guard, 404 two seconds into the assertion window without it.

No `For()` predicate here, per your point that it cannot cover the map-func enqueues.

johannes-engler-mw:
Rebased on master; CI was red for two reasons, one of them mine.

**Mine:** #2839 dropped the `gatewayv1alpha2` import from the L4 controllers, and the delete
blocks here still used `gatewayv1alpha2.GroupVersion`. Beyond the build failure that would have
been wrong at runtime — `label.GenLabel` derives the delete labels from that Kind/APIVersion, so
the delete would no longer match what the update path writes. Now `gatewayv1.GroupVersion`,
matching the sibling block above it.

**Not mine:** `recordingProvider` is declared twice on master — `apisixconsumer_controller_test.go:45`
(#2850) and `gateway_controller_publishservice_test.go:49` (#2846). `internal/controller` fails
`go vet` on master alone; I checked a clean worktree with nothing of this branch in it. That fails
`run-test` and `lint` for every open PR, and stops the e2e jobs at the image build.

`096d47c7` fixes it by keeping the ApisixConsumer stub and giving it the `Update` counter the
Gateway test needs. It is first on the branch and touches nothing else, so it can be cherry-picked
on its own — happy to move it to a separate PR if you would rather.

Re-verified locally after the rebase: `go vet` and the unit suite clean, and the
`HTTPRoute with Multiple Gateway` e2e context passes on kind (3/3).

johannes-engler-mw:
Moved the stub fix out to #2854, since it is a master build break rather than anything to do with
this change. It stays on this branch for now so the tests here can run; it will drop out on the
next rebase once #2854 lands.

johannes-engler-mw:
Ping @AlinsRan

AlinsRan (approved):
LGTM

nic-6443 (approved):

PR #2838, opened 2026-08-07
chore: v2.2.0 release

AlinsRan:
## Description

Release PR for **apisix-ingress-controller 2.2.0**, covering the 32 commits merged since `2.1.0`.

Following the process in `releases/README.md`, this PR only prepares the release notes and version:

- `releases/v2.2.0.toml` — release-tool template for the new version (`previous = "2.1.0"`, `pre_release = false`)
- `CHANGELOG.md` — 2.2.0 section (highlights, upgrade notes, breaking changes, features with examples, security/bug fixes, chores, docs, contributors, commit list, dependency changes) and the TOC entry
- `Makefile` — `VERSION ?= 2.1.0` -> `2.2.0`

## Highlights

- Gateway API 1.6.0 (#2804); L4 routes and ReferenceGrant read through `gateway.networking.k8s.io/v1` (#2839)
- Downstream mTLS on Gateway listeners at `spec.tls.frontend` (#2792)
- `L4RoutePolicy` for attaching stream plugins to L4 routes (#2791)
- L4 upstream schemes `tcp`/`tls`/`udp` (#2830)
- Security: credential/AdminKey redaction in logs (#2808), ReferenceGrant on cross-namespace Consumer SecretRef (#2805), no cross-namespace Secret existence oracle in the Consumer webhook (#2806)

## Upgrade notes and breaking changes

The notes separate what merely needs attention from what actually breaks.

**Upgrade notes** (not breaking): Kubernetes 1.31+ is *recommended* — `checkK8sVersion` only logs a warning on older clusters, the controller still starts. The real 1.31 constraint is on the Gateway API 1.6.0 CRDs, which also have to be installed with `kubectl apply --server-side --force-conflicts`.

**Breaking changes** — three, each with the config change spelled out in the notes:

1. Cross-namespace `Consumer` `secretRef` now needs a `ReferenceGrant` (#2805) — example grant included; without it the `Consumer` stops reconciling.
2. L4 route webhook paths move from `/validate-gateway-networking-k8s-io-v1alpha2-*` to `-v1-*` (#2839); hand-maintained `ValidatingWebhookConfiguration`s need updating.
3. `enable-csrf` without `csrf-key` is rejected by the webhook (#2813).

#2818 is **not** a breaking change and has been moved to the upgrade notes: the `server_port` injection it adds for `TCPRoute`/`UDPRoute` is gated by `listener_port_match_mode`, which defaults to `off`. The notes carry a mode table plus the listener-port requirement that applies once it is turned on.

Downstream mTLS (#2792) is documented with a full Gateway example, since Gateway API 1.6 declares `frontendValidation` at `spec.tls.frontend` rather than on the listener. `L4RoutePolicy` and the new L4 `scheme` values have examples too. CRD diff since 2.1.0 was checked for renames or removals: none, the changes are additive (`scheme` enum extended, `L4RoutePolicy` added).

## Follow-ups (not in this PR)

- `docs/en/latest/overview.md` compatibility table needs a `2.2.0` row, and `docs/en/latest/getting-started/configure-routes.md` still points at the `v2.1.0` branch.
- `docs/en/latest/upgrade-guide.md` only covers 1.x -> 2.0.0; a 2.1.0 -> 2.2.0 section would be worth adding.

As with 2.1.0 (#2776), these go in a separate docs PR after the tag exists.

## Checklist

- [x] I have explained the need for this PR and the problem it solves
- [x] I have explained the changes or the new features added to this PR
- [ ] I have added tests corresponding to this change
- [x] I have updated the documentation to reflect this change
- [x] I have verified that this change is backward compatible (if not, please discuss on the APISIX mailing list first)


nic-6443 (approved):

releases/v2.2.0.toml
```diff
@@ -0,0 +1,268 @@
...
+* **Downstream mTLS** on `Gateway` listeners, configured at `spec.tls.frontend` [#2792](https://github.com/apache/apisix-ingress-controller/pull/2792).
+* **`L4RoutePolicy`**, a new CRD that attaches stream plugins to L4 routes [#2791](https://github.com/apache/apisix-ingress-controller/pull/2791).
+* **L4 upstream schemes** `tcp`, `tls` and `udp` for stream backends [#2830](https://github.com/apache/apisix-ingress-controller/pull/2830).
+* **Security fixes**: credentials are kept out of the logs, a cross-namespace `Consumer` `secretRef` is gated by a `ReferenceGrant`, and the `Consumer` webhook no longer reveals whether a `Secret` exists in another namespace.
+
+---
+
+## Upgrade Notes
+
+**Kubernetes 1.31 or newer is recommended.** On an older cluster the controller logs a
```
kayx23: `1.31` is the declared minimum supported version, not merely a recommendation: `MIN_K8S_VERSION` is `1.31.0`, and both `README.md` and `docs/en/latest/install.md` say 1.31+ is required. Continuing after the startup warning does not make older clusters supported, so this wording could lead operators into an unsupported upgrade. Please state 1.31+ as the minimum supported/required version here and in the generated changelog; the warning-only enforcement can be noted separately.

releases/v2.2.0.toml
```diff
@@ -0,0 +1,268 @@
...
+`UDPRoute` join `HTTPRoute` and `GRPCRoute` under the `server_port` route-var injection
+this option gates. The default `off` keeps the previous behavior: one portless
+`StreamRoute` per rule, matching every connection on the stream listener. The other two
+modes now apply to L4 as well:
+
+| mode | when `server_port` is injected |
+| ---- | ------------------------------ |
+| `off` (default) | never |
+| `explicit` | a parentRef pins `sectionName` or `port` |
+| `auto` | that, or the `Gateway` has more than one listener port |
```
kayx23: `auto` checks the distinct listener ports matched by this route (`tctx.Listeners`), not every listener configured on the Gateway. An unrelated second Gateway listener therefore does not trigger injection. Please change this to “the route attaches to more than one listener port” here and in `CHANGELOG.md`.

releases/v2.2.0.toml
```diff
@@ -0,0 +1,268 @@
...
+  name: consumer-to-secret
+  namespace: secret-ns          # where the Secret lives
+spec:
+  from:
+    - group: apisix.apache.org
+      kind: Consumer
+      namespace: consumer-ns    # where the Consumer lives
+  to:
+    - group: ""
+      kind: Secret
```
kayx23, line 112: Without `to.name`, this grant authorizes every `Consumer` in `consumer-ns` to reference every `Secret` in `secret-ns`, which is broader than the credential migration requires. Please scope the example to the credential Secret (for example, `name: consumer-credentials`) or explicitly warn that omitting the name grants access to all Secrets of this kind.

releases/v2.2.0.toml
AlinsRan: Adopted the "supported version" framing, but not "required": I could not substantiate a hard 1.31 gate. Gateway API 1.6.0 states no Kubernetes minimum of its own, its CRDs use no CEL feature newer than the 1.30-era set (no `optionalOldSelf`), the standard-channel ValidatingAdmissionPolicy is `admissionregistration.k8s.io/v1` (GA in 1.30), and this repo still runs envtest at `ENVTEST_K8S_VERSION = 1.30.0`. `checkK8sVersion` only logs.

The notes now read: "Kubernetes 1.31+ is the supported version for this release, up from 1.26.0, as declared by `MIN_K8S_VERSION`, `README.md` and the installation guide" followed by "This is a support boundary rather than a hard gate: on an older cluster the controller logs a warning and keeps running. Below 1.31 you are outside what the project tests and supports." That matches the declared minimum without telling operators an older cluster cannot run.

Separately, the rationale in `install.md` looks overstated (IngressClass namespaced params are GA since 1.23, EndpointSlice terminating conditions since 1.26); worth revisiting in the docs PR.

releases/v2.2.0.toml
AlinsRan: Correct. `listenerPortSet` builds the set from `tctx.Listeners`, which the L4 controllers populate only from the parents `ParseRouteParentRefs` matched, so an unrelated listener on the same Gateway does not count. Reworded to "the route attaches to more than one listener port" in both files.

releases/v2.2.0.toml
AlinsRan: Right, `checkReferenceGrant` treats a nil `to.name` as a wildcard (`to.Name == nil || *to.Name == ref.Name`). The example now sets `name: consumer-credentials` with a trailing comment that omitting it grants every Secret in the namespace.

releases/v2.2.0.toml
kayx23: Verified on `0fd98043`: the release template and generated changelog now state 1.31+ as the supported boundary while accurately distinguishing the warning-only runtime check. This resolves the concern.

releases/v2.2.0.toml
kayx23: Verified on `0fd98043`: both files now describe `auto` in terms of listener ports matched by the route, consistent with `tctx.Listeners` and `listenerPortSet`. This resolves the concern.

releases/v2.2.0.toml
kayx23: Verified on `0fd98043`: the example now scopes `to.name` to `consumer-credentials` and explicitly explains the wildcard effect of omission in both files. This resolves the concern.

bzp2010:
BTW, once this is done, please don't forget to submit a compatibility report to the gateway-api project.

kayx23 (approved):

shreemaan-abhishek (approved):

PR #2846, opened 2026-08-12
fix: fall back to publishService for Gateway status addresses

### LINKED ISSUE #2800: Gateway status.addresses should fall back to publishService like the Ingress path
### Type

Feature request / behavior inconsistency between the Ingress and Gateway API reconcilers.

### Current Behavior

In `2.1.0`, the **Gateway API** controller populates `Gateway.status.addresses` **only** from `gatewayProxy.spec.statusAddress`. There is no fallback to `publishService`.

`internal/controller/gateway_controller.go`:
```go
gatewayProxy, ok := tctx.GatewayProxies[rk]
if !ok {
    acceptStatus = conditionStatus{status: false, msg: "gateway proxy not found"}
} else {
    for _, addr := range gatewayProxy.Spec.StatusAddress {   // only StatusAddress
        addrType := gatewayv1.IPAddressType
        if net.ParseIP(addr) == nil { addrType = gatewayv1.HostnameAddressType }
        addrs = append(addrs, gatewayv1.GatewayStatusAddress{Type: &addrType, Value: addr})
    }
}
...
gateway.Status.Addresses = addrs   // overwrites; if StatusAddress is empty, addresses are cleared on every reconcile
```

By contrast, the **Ingress** controller already resolves `publishService` dynamically when `statusAddress` is not set.

`internal/controller/ingress_controller.go` (`updateStatus`):
```go
// 1. use the IngressStatusAddress in the config
statusAddresses := gatewayProxy.Spec.StatusAddress
if len(statusAddresses) > 0 {
    ... // static
} else {
    // 2. if the IngressStatusAddress is not configured, try to use the PublishService
    publishService := gatewayProxy.Spec.PublishService
    if publishService != "" {
        ... // resolves the LoadBalancer Service ingress IP/hostname dynamically
    }
}
```

So `publishService` works for Ingress status but is ignored for Gateway status.

### Expected Behavior

The Gateway reconciler should mirror the Ingress path: when `gatewayProxy.spec.statusAddress` is empty, fall back to `gatewayProxy.spec.publishService` and populate `Gateway.status.addresses` from that Service's `.status.loadBalancer.ingress[].ip|hostname`.

### Why

On cloud LoadBalancers (we run OpenStack Octavia; AWS/GCP are equivalent), the external IP/hostname is assigned dynamically and can change when the LB is recreated. A static `statusAddress` is brittle: it must be hand-updated whenever the LB address changes, and if left empty the controller clears `Gateway.status.addresses` on every reconcile.

This breaks consumers that read `Gateway.status.addresses` — e.g. external-dns (`--source=gateway-httproute`) publishes A records from the Gateway address; an empty address causes it to remove all records for routes attached to the Gateway. `publishService` already exists on the GatewayProxy and is the natural dynamic source; it just isn't wired into the Gateway status path.

### Environment

- apisix-ingress-controller `2.1.0` (manager), ADC `0.26.0`
- Gateway API standalone mode, `gatewayProxy.spec.publishService` set to the `LoadBalancer` Service, `statusAddress` unset
- LoadBalancer: OpenStack Octavia (dynamic VIP)

### Proposed Change

In `gateway_controller.go`, after the `StatusAddress` loop, when no `statusAddress` entries are present, resolve `gatewayProxy.spec.publishService` to the Service's LoadBalancer ingress and append the resulting `GatewayStatusAddress` entries (IP → `IPAddress`, hostname → `Hostname`), reusing the helper logic already used by the Ingress path. Happy to contribute a PR if the direction is acceptable.

Related prior work: #2730 / #2732 (statusAddress hostname support + publishService/statusAddress logic).

janiussyafiq:
### Type of change:

- [x] Bugfix

Fixes #2800.

`Gateway.status.addresses` was only populated from `GatewayProxy.spec.statusAddress`.
When `spec.publishService` is set instead, the Gateway published no address, even though the Ingress status path already falls back to the publish Service's LoadBalancer address.

This PR extracts shared publish-Service resolution helpers, makes `Gateway.status.addresses` fall back to the LoadBalancer address of `spec.publishService` when `statusAddress` is empty (`statusAddress` still wins when both are set), and refactors the Ingress LoadBalancer status branch onto the same helpers.

Note: addresses are typed with `net.ParseIP` (existing convention), so a LoadBalancer ingress entry whose `hostname` field holds a literal IP string is published as an IP-type address.

A follow-up PR will add a `publishService` field index and a Service watch on both reconcilers, so a reassigned LoadBalancer address re-triggers status reconciliation instead of waiting for an unrelated event.

### Pre-submission checklist:


AlinsRan (approved):

nic-6443 (approved):

internal/controller/gateway_controller.go
```diff
@@ -194,10 +194,14 @@ func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
-		for _, addr := range gatewayProxy.Spec.StatusAddress {
-			if addr == "" {
-				continue
-			}
+		statusAddresses, err := r.resolveStatusAddresses(ctx, &gatewayProxy)
+		if err != nil {
+			// fail the reconcile so a missing or invalid publish Service retries
+			// with backoff, mirroring the Ingress status path
+			r.Log.Error(err, "failed to resolve gateway status addresses", "gateway", req.NamespacedName)
+			return ctrl.Result{}, err
```
AlinsRan: This returns before `r.Provider.Update` (L234), so a `publishService` that cannot be resolved — a typo, or the Service not created yet — stops the Gateway from being pushed to APISIX at all. A status-only problem becomes a data plane outage.

internal/controller/gateway_controller.go
```diff
@@ -259,6 +263,38 @@ func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+			return addr != ""
+		}), nil
+	}
+
+	if gatewayProxy.Spec.PublishService == "" {
+		return nil, nil
+	}
+
+	// a bare name is resolved against the GatewayProxy's namespace
+	svc, err := resolvePublishService(ctx, r.Client, gatewayProxy.Spec.PublishService, gatewayProxy.GetNamespace())
```
AlinsRan: The doc comment above claims both APIs end up with the same addresses, but for a bare name (`publishService: apisix-gateway`, no namespace) they don't — the two paths default to a different namespace:

- here: the GatewayProxy's namespace
- Ingress path: `ingress.Namespace` (ingress_controller.go:716)

Those agree only when the Ingress happens to sit in the GatewayProxy's namespace. With the GatewayProxy in `ingress-apisix` and the Ingress in `app`, the Ingress path looks up `app/apisix-gateway` and fails with NotFound, while this path looks up `ingress-apisix/apisix-gateway` and succeeds.

The rule used here is the better one — suggest making the Ingress path use it too, and dropping the "same addresses for both APIs" claim from the comment.

AlinsRan (changes requested):

internal/controller/gateway_controller.go
janiussyafiq: fixed

internal/controller/gateway_controller.go
janiussyafiq: fixed

internal/controller/gateway_controller.go
```diff
@@ -253,10 +258,41 @@ func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 			}),
 		})
 
-		return ctrl.Result{}, nil
+		return ctrl.Result{}, addrResolveErr
+	}
+
+	return ctrl.Result{}, addrResolveErr
```
AlinsRan: The data plane is no longer blocked, thanks. Returning the error still has two costs worth weighing:

- A permanent failure retries forever. `publishService: a/b/c` fails in `SplitMetaNamespaceKey`, and a typo'd name is NotFound until someone edits the GatewayProxy — neither can ever succeed on retry. With the default rate limiter that is one full reconcile every 1000s, `r.Provider.Update` included, so a typo keeps re-translating and re-pushing to APISIX indefinitely.
- It shows up as a controller failure. Every attempt is logged at error level and counted in `controller_runtime_reconcile_errors_total`, so a status-address problem trips alerts built on that metric.

AlinsRan (approved):

PR #2856, opened 2026-09-03
feat: report dp instance unavailable to gateway proxy

bzp2010:
### Type of change:

- [ ] Bugfix
- [x] New feature provided
- [ ] Improve performance
- [ ] Backport patches
- [ ] Documentation
- [x] Refactor
- [ ] Chore
- [ ] CI/CD or Tests

Accurately report data plane sync failures, at both the GatewayProxy level and the individual resource level.

#### GatewayProxy: `DataPlaneAvailable` condition

The `GatewayProxy` that addresses the affected instances gets a new `status.conditions` field (standard Kubernetes conditions, not a custom shape), reporting two kinds of failure:

- **apisix-standalone**: per-instance failures reported by ADC's `endpoint_status`, whether a specific instance is unreachable, rejects the write, or a mix. This tracks whether each instance took the write, not which resource caused a rejection, so it's reported here regardless of whether the same sync also named a specific resource (see Resources below): a rejected write leaves every instance failed in `endpoint_status` even on the runs where ADC's re-validate step also manages to pin the cause on one resource, and both get reported, not one in place of the other.
- **Both backends**: a hard error from ADC itself (e.g. every server unreachable, so ADC never got far enough to attempt any resource) has no structured per-resource body to attribute at all.

The condition is recomputed from scratch every sync round, directly from that round's actual result, never diffed against remembered state, so a GatewayProxy that has always been healthy still gets an explicit `True` the first time it syncs, and a restart can never leave a stale `False` stuck behind.

```yaml
# bad
status:
  conditions:
  - type: DataPlaneAvailable
    status: "False"
    observedGeneration: 1
    lastTransitionTime: "2026-09-04T02:10:03Z"
    reason: DataPlaneInstanceUnavailable
    message: "1/2 gateway instance(s) failed to apply the last sync: http://unreachable-instance.invalid:9180: dial tcp: lookup unreachable-instance.invalid: server misbehaving"

# good
status:
  conditions:
  - type: DataPlaneAvailable
    status: "True"
    observedGeneration: 1
    lastTransitionTime: "2026-09-04T02:25:11Z"
    reason: DataPlaneAvailable
    message: ""
```

Per-instance failures are also reported as `Warning` events, one per failed endpoint, so each instance's own failure history (when it started, how often) stays visible on its own instead of folding into one aggregated message. The event's `involvedObject` carries the GatewayProxy's real UID, so it shows up under `kubectl describe gatewayproxy`.

```
Events:
  Type     Reason                        Age                 From             Message
  ----     ------                        ----                ----             -------
  Warning  DataPlaneInstanceUnavailable  8s (x23 over 12m)   apisix-provider  http://apisix-2:9180: dial tcp 10.244.0.12:9180: connect: connection refused
  Warning  DataPlaneInstanceUnavailable  8s (x23 over 12m)   apisix-provider  https://apisix-3:9180: tls: handshake failure
```

```yaml
apiVersion: v1
kind: Event
metadata:
  name: gateway-proxy-partial-instance.17f2a3c1d0e5f6a7
  namespace: ns-e2e-test
count: 23
firstTimestamp: "2026-09-04T02:10:03Z"
lastTimestamp: "2026-09-04T02:22:11Z"
type: Warning
reason: DataPlaneInstanceUnavailable
message: "http://apisix-2:9180: dial tcp 10.244.0.12:9180: connect: connection refused"
involvedObject:
  kind: GatewayProxy
  name: gateway-proxy-partial-instance
  namespace: ns-e2e-test
source:
  component: apisix-provider
```

#### Resources: `Accepted`/`SyncFailed` condition

A sync failure ADC *can* attribute to a specific resource (a rejected route, a bad plugin config, or, for the `apisix` backend, a transient network error while pushing that one resource) is reported on that resource's own status instead, for both backends. Unlike GatewayProxy, this keeps a small persisted delta (a config's resource set can be large, so recomputing all of it every round would be wasteful): newly or still-failing resources are written `SyncFailed`, and any resource that stops failing gets its error explicitly cleared with an `Accepted` write.

```yaml
status:
  conditions:
  - type: Accepted
    status: "False"
    reason: SyncFailed
    message: "ServerAddr: http://apisix:9180, Error: unknown plugin [non-existent-plugin]"
```

As noted above, this and the GatewayProxy's `endpoint_status` reporting aren't alternatives; the same sync can produce both at once.

This part is expected to improve further: apisix-standalone is planned to gain a process that automatically excludes the resource ADC named as bad and retries the rest, so the sync self-heals instead of leaving the whole write rejected until someone fixes the offending resource by hand.

#### ADC version

This depends on ADC >= 0.30.3: earlier 0.30.x releases serialized the sync response's per-resource `event` envelope (`resourceType`/`resourceId`/...) with the wrong key casing, which silently broke resource-level attribution. `.github/workflows/apisix-e2e-test.yml` and the default sidecar image in `config/manager/kustomization.yaml` are both bumped to 0.30.3.

### Pre-submission checklist:


bzp2010:
There is an issue that needs to be discussed:

Currently, this PR's change is limited to apisix-standalone, while apisix still propagates errors to resources—whether the Admin API is unavailable or there is an error in the configuration itself. Should we distinguish between these cases?

AlinsRan:
> There is an issue that needs to be discussed:
> 
> Currently, this PR's change is limited to apisix-standalone, while apisix still propagates errors to resources—whether the Admin API is unavailable or there is an error in the configuration itself. Should we distinguish between these cases?

Sync errors should be recorded on the route resources — that status is how users check whether a route has actually taken effect.

For `GatewayProxy`, it may make more sense to record only Admin API unavailability.

internal/provider/apisix/status.go
```diff
@@ -42,32 +55,101 @@ import (
...
+	}
+}
+
 //nolint:gocyclo
 func (d *apisixProvider) updateStatus(nnk types.NamespacedNameKind, condition metav1.Condition) {
 	switch nnk.Kind {
+	case types.KindGatewayProxy:
+		// Unlike the route kinds below, the condition lands on the GatewayProxy's own
+		// top-level Status.Conditions, not on a per-parent entry.
+		d.updater.Update(status.Update{
```
AlinsRan, line 240: This routes a new Kind through `UpdateHandler` that `statusEqual` doesn't know about. `internal/controller/status/updater.go:179` has no `*v1alpha1.GatewayProxy` case, so it falls through to `default: return false` — two identical conditions never compare equal, and the dedup shortcut at `updater.go:113` never fires.

Verified with a probe test (control included):

```
statusEqual(GatewayProxy{c}, GatewayProxy{c}) == false   // no case -> default
statusEqual(Consumer{},      Consumer{})      == true    // has a case
```

`newGatewayProxyDataPlaneAvailableCondition` also stamps a fresh `metav1.Now()` every cycle, so the serialized object differs too and the apiserver's no-op-write optimization doesn't save us either. Every sync issues a real `Status().Update()` and bumps `resourceVersion`.

That write then fans out. `gateway_controller.go` is the only one of the 13 GatewayProxy watchers without a `WithEventFilter`, and its `Watches(&v1alpha1.GatewayProxy{}, ...)` carries no predicate either, so a status-only update enqueues every Gateway whose `parametersRef` points here. `Reconcile` unconditionally calls `Provider.Update` (`gateway_controller.go:259`), which reaches `defer d.syncNotify()` (`provider.go:169`). `syncNotify` is an independent wake source in the sync loop's `select`, so it bypasses the retrier completely:

```
GatewayProxy status write -> Gateway reconcile -> syncNotify -> sync fails -> status write -> ...
```

While an endpoint stays down, this turns what should be a backoff (`RetryMaxDelay` 1000s, or the 1h ticker) into a continuous full resync: healthy instances get the entire config rewritten every cycle, and `failed to run http sync for server` floods the log. It stops on its own once the data plane recovers, but not before.

Adding a `case *v1alpha1.GatewayProxy` to `statusEqual` breaks the cycle at the source — `cmpIgnoreLastTT` already ignores `LastTransitionTime`, so an unchanged condition would be skipped. Giving the `gateway_controller` GatewayProxy watch a `GenerationChangedPredicate` would be a good second line of defence, though that one is pre-existing rather than introduced here.

internal/provider/apisix/status.go
```diff
@@ -42,32 +55,101 @@ import (
...
 	// Mark resources that exist only in the previous failure history as successful.
 	for nnk := range d.statusUpdateMap {
 		if _, ok := statusUpdateMap[nnk]; !ok {
-			d.updateStatus(nnk, cutils.NewConditionTypeAccepted(
-				apiv2.ConditionReasonAccepted,
-				true,
-				0,
-				"",
-			))
+			d.updateStatus(nnk, successCondition(nnk))
```
AlinsRan: Recovery is inferred from "the key left the failure set", not from "this sync succeeded".
ADC produces the former without the latter:

| ADC `/sync` response | carries `endpoint_status`? |
| --- | --- |
| 202 success | yes (apisix-standalone) |
| 400 schema/lint, 500 catch-all | **no** |

Checked at `api7/adc` main (`9b0e093`) — same in both the Node and Rust servers.

So when ADC itself fails, `unavailableEndpointsMessage` returns `""` (status.go:368), the GatewayProxy never enters `statusUpdateMap`, and this branch flips it to `True` plus a "every data plane instance took the last sync" event — while the sync is still failing. The harder ADC fails, the greener the GatewayProxy gets.

Recovery should be driven by an actually-successful sync, with "ADC Server itself failed" leaving the condition untouched.

(Possibly related to the failing e2e you linked: on the Node ADC, `findLatest()` has no per-server error handling, so one unreachable instance throws out of `dump()` before `sync()` runs — 500, no `endpoint_status`.)

api/v1alpha1/gatewayproxy_types.go
```diff
@@ -178,6 +179,19 @@ type GatewayProxy struct {
 	// GatewayProxySpec defines configuration of gateway proxy instances,
 	// including networking settings, global plugins, and plugin metadata.
 	Spec GatewayProxySpec `json:"spec,omitempty"`
+
+	// Status defines the current state of Gateway Proxy.
+	//
+	// +kubebuilder:default={conditions: {{type: "DataPlaneAvailable", status: "Unknown", reason:"Pending", message:"Waiting for controller", lastTransitionTime: "1970-01-01T00:00:00Z"}}}
```
AlinsRan, line 185: This default is never overwritten on the happy path — an always-healthy GatewayProxy reports `Waiting for controller` forever. That contradicts the "good" example in the PR description, where `status: "True"` is shown as the healthy steady state.

Cause: `handleStatusUpdate` writes the success condition only for keys that were in the *previous* failure set, and nothing else ever writes `DataPlaneAvailable`. Two effects:

- healthy from the start never turns `True`
- a stale `False` survives a restart — `d.statusUpdateMap` is in-memory, so the success branch never fires after a restart

internal/provider/apisix/status.go
```diff
@@ -42,32 +55,101 @@ import (
...
+}
+
+// recordFailedEndpointEvents fires one Warning event per failed EndpointStatus entry,
+// so each instance's own failure history (when it started, how often) is visible on
+// its own, not folded into everyone else's.
+func (d *apisixProvider) recordFailedEndpointEvents(nnk types.NamespacedNameKind, endpoints []adctypes.EndpointStatus) {
+	if d.EventRecorder == nil {
+		return
+	}
+	gatewayProxy := &apiv1alpha1.GatewayProxy{
```
AlinsRan: This stub has no UID, so the Event's `involvedObject.uid` is empty and `kubectl describe gatewayproxy` won't show it — which is what the new troubleshooting section tells users to run: [troubleshoot.md#L77-L88](https://github.com/apache/apisix-ingress-controller/blob/6d9930834abc81efc59719d8846b5f69ece03695/docs/en/latest/reference/troubleshoot.md#L77-L88)

`client-go` adds `involvedObject.uid=<uid>` to the field selector whenever the described object has a UID, which a real GatewayProxy always does. Verified on a live cluster: same object and event, without the uid `describe` prints `Events: <none>`, with it the event shows. The e2e passes because it selects on `kind` + `name` only.

Either record against the real object from the informer cache (needs `mgr.GetClient()` threaded through from `run.go`), or change that doc block to `kubectl get events --field-selector ...`.

internal/provider/apisix/status.go
bzp2010: fixed

internal/provider/apisix/status.go
bzp2010: fixed

api/v1alpha1/gatewayproxy_types.go
bzp2010: fixed

internal/provider/apisix/status.go
bzp2010: fixed

internal/provider/apisix/status.go
bzp2010: Confirmed `statusEqual` already special-cases GatewayProxy (see `internal/controller/status/updater.go`), so writes there dedupe the same way every other kind's do. No further change needed for this one.

internal/provider/apisix/status.go
bzp2010: Reworked the whole status pipeline: GatewayProxy's DataPlaneAvailable is now recomputed from scratch every sync round straight from that round's actual result, never diffed against remembered history. A round with no error writes True, any error writes False — recovery is never inferred from a key merely leaving a failure set. This makes the class of bug you describe structurally impossible rather than patching the specific case. See `updateStatusFromSyncResults`/`classifySyncResult` in `internal/provider/apisix/status.go`.

Cost note, since this write path is now unconditional every round instead of only on change: the enqueue itself is cheap (a channel send), and the actual write is still deduped by the pre-existing `statusEqual` check in `UpdateHandler.updateStatus` (`internal/controller/status/updater.go`), so an unchanged GatewayProxy never triggers a real PATCH. The `client.Get` that check needs reads through the manager's shared informer cache, not the API server — GatewayProxy already has its own controller/watch (`GatewayProxyController`), so this isn't a fresh network round trip. The added cost is bounded to the (typically small) number of GatewayProxies; the per-resource path (ApisixRoute etc.) is untouched and keeps the old change-only write behavior.

api/v1alpha1/gatewayproxy_types.go
bzp2010: Same fix as the other GatewayProxy status comment: the condition is now written unconditionally every round based on that round's actual outcome, with no dependency on any remembered history. A GatewayProxy that's always been healthy gets True the first time it syncs, and a restart can no longer leave a stale False stuck forever, since nothing carries over between rounds for GatewayProxy specifically.

internal/provider/apisix/status.go
bzp2010: recordFailedEndpointEvents now fetches the real GatewayProxy via the client before firing the Event, instead of building a bare Name/Namespace stub, so involvedObject carries the real UID and shows up under `kubectl describe gatewayproxy`.

internal/provider/apisix/status.go
```diff
@@ -18,56 +18,232 @@
...
+	case types.KindGatewayProxy:
+		// Unlike the route kinds below, the condition lands on the GatewayProxy's own
+		// top-level Status.Conditions, not on a per-parent entry.
+		d.updater.Update(status.Update{
+			NamespacedName: nnk.NamespacedName(),
+			Resource:       &apiv1alpha1.GatewayProxy{},
+			Mutator: status.MutatorFunc(func(obj client.Object) client.Object {
+				cp := obj.(*apiv1alpha1.GatewayProxy).DeepCopy()
+				condition.ObservedGeneration = cp.GetGeneration()
+				cp.Status.Conditions = cutils.MergeCondition(cp.Status.Conditions, condition)
```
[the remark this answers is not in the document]
bzp2010: This will be fixed in a future PR; it's a completely unrelated issue.

nic-6443 (approved):

shreemaan-abhishek (approved):

AlinsRan (approved):

PR #2874, opened 2026-09-15
fix: remove consumer config after ingress class handoff

shreemaan-abhishek:
### Type of change:

- [x] Bugfix

Retracts previously generated consumer configuration when an `ApisixConsumer` switches to an IngressClass that this controller does not manage, or when its selected class is removed.

Transient Kubernetes lookup failures are returned for retry so existing configuration is retained until class ownership can be determined reliably. Unit and end-to-end coverage verifies class handoff, selected-class removal, and transient-error behavior. Provider deletion logs now include only the Kubernetes object identity.

Validated with:

- `go test ./internal/controller ./internal/provider/apisix -count=1`
- `go test ./test/e2e/apisix -run '^$'`
- `make lint`
- `make test`

### Pre-submission checklist:


AlinsRan:
Need e2e test.

shreemaan-abhishek:
Added end-to-end coverage for both class handoff and deletion of the selected class. Each case first confirms the key works, then verifies it stops succeeding and that the consumer is absent from APISIX.

AlinsRan (approved):

nic-6443 (approved):

PR #2875, opened 2026-09-15
fix: return an error response for unresolved ExtensionRef

### LINKED ISSUE #2841: HTTPRoute/GRPCRoute: unresolvable ExtensionRef plugin config should return a data-plane error, not skip the filter or drop the route
### Problem

When an `ExtensionRef` filter on an `HTTPRoute`/`GRPCRoute` points at a `PluginConfig` whose plugin `config` cannot be resolved, `fillPluginFromExtensionRef` currently skips the plugin, and the route is published without it.

Gateway API forbids that outcome. A custom filter that cannot be resolved must not be skipped; the requests that would have been processed by it must receive an HTTP error response instead:

> If a reference to a custom filter type cannot be resolved, the filter MUST NOT be skipped. Instead, requests that would have been processed by that filter MUST receive a HTTP error response.

(gateway-api v1.6.0, `apis/v1/httproute_types.go#L847-L849`)

Failing translation outright is not the fix either. It means the route is never programmed, so if another route also matches, the request is served **without** the filter — the exact outcome the spec forbids — and if nothing else matches it is a 404, indistinguishable from "no such route". It also widens a per-filter failure into a whole-route one: `TranslateHTTPRoute` aborts on the first bad filter, dropping every rule of the route, and one bad `PluginConfig` takes down every route referencing it.

### Expected behavior

Adopt the conformant pattern the translator already uses for unresolvable `backendRefs`: keep the route, and attach a `fault-injection` plugin returning 500 so the affected requests fail at the data plane rather than silently bypassing the filter. See `internal/adc/translator/httproute.go` (the `backendErr` handling and its `fault-injection` abort block), whose error is deliberately discarded at the call site precisely because it has already been turned into a data-plane 500.

Scope the failure to the rule whose filter could not be resolved, not the whole route, and set `ResolvedRefs=False` on the route so the condition is visible.

### Affected

- `internal/adc/translator/httproute.go` — `fillPluginFromExtensionRef`, `fillPluginsFromHTTPRouteFilters`
- `internal/adc/translator/grpcroute.go` — same `ExtensionRef` path

### Context

Came out of review on #2814, which makes the apiv2 plugin-config paths fail closed. The Gateway API paths were deliberately left out of that PR because they need this different, spec-conformant treatment. The apiv2 CRDs carry no such constraint.

Depends on #2840 for the failure to be observable in status at all.

shreemaan-abhishek:
### Type of change:

- [x] Bugfix

This change validates both the group and kind of Gateway API ExtensionRef filters. The supported reference is `apisix.apache.org/PluginConfig`.

Unsupported group or kind values now set `ResolvedRefs=False` with reason `InvalidKind`. Missing PluginConfig or referenced Secret objects set `ResolvedRefs=False` with reason `BackendNotFound`.

HTTPRoute and GRPCRoute translation now keeps the route active and adds a per-rule `fault-injection` response with status 500 when an ExtensionRef cannot be resolved. Other rules on the same route remain active.

Unit coverage includes reference validation, status conditions, HTTPRoute and GRPCRoute translation, and ExtensionRef indexers. An end-to-end HTTPRoute case verifies both the condition and the 500 response.

Fixes #2841

### Breaking changes

Rules with invalid or unresolved ExtensionRef filters previously forwarded requests without the requested filter. Those rules now return 500. Use `apisix.apache.org/PluginConfig` and ensure the referenced PluginConfig and Secrets exist. Valid ExtensionRef behavior is unchanged.

### Pre-submission checklist:


internal/adc/translator/httproute.go
```diff
@@ -711,7 +723,7 @@ func (t *Translator) TranslateHTTPRoute(tctx *provider.TranslateContext, httpRou
 		enableWebsocket, _ := t.translateBackendsToUpstreams(tctx, rule, httpRoute, service)
 
 		if err := t.fillPluginsFromHTTPRouteFilters(service.Plugins, httpRoute.GetNamespace(), rule.Filters, rule.Matches, tctx); err != nil {
-			return nil, err
+			setExtensionRefErrorResponse(service)
```
AlinsRan, line 731: The error log is missing here. It will hinder the debug.

shreemaan-abhishek: Fixed in 73774fb6. HTTPRoute and GRPCRoute translation now log the original filter error with the route namespace, name, and rule index before installing the per-rule error response. Added regression coverage, including an assertion that configuration content is not logged.

AlinsRan (approved):

nic-6443 (approved):

PR #2882, opened 2026-09-17
feat: serve TLSRoute in Passthrough mode

AlinsRan:
## What this does

APISIX's stream proxy can now forward a TLS stream to the upstream untouched while still picking that upstream from the SNI, which it prereads from the `ClientHello` rather than learning from a handshake it performed itself (apache/apisix#13912). A stream listen opts in per port:

```yaml
apisix:
  stream_proxy:
    tcp:
      - addr: 9110
        tls: true              # terminate (unchanged)
      - addr: 9120
        tls_passthrough: true  # pass everything through
```

That is exactly what Gateway API asks of a listener in `Passthrough` mode, and it is the one thing the stream subsystem could not do before: routing by SNI implied terminating.

So a `Passthrough` listener behaved as `Terminate`. The translator never looked at `tls.mode`, the gateway decrypted a stream the backend was supposed to own, and the handshake failed against a certificate the gateway does not have.

- `TranslateTLSRoute` reads the `tls.mode` of the listeners the route attached to and sets `tls_passthrough` on the stream routes bound to a `Passthrough` port. The TLSRoute reconciler populates `tctx.Listeners` for that, which the TCPRoute and UDPRoute ones already did.
- Every matched listener on a port has to agree. A physical stream listen is either terminating or prereading, never both, so listeners that disagree fall back to terminating rather than to a guess. Within one Gateway such a port is already `ProtocolConflict` and attaches no routes at all, so this only concerns a port two different Gateways disagree about.

## Conformance

The four TLSRoute tests pinned to `mode: Passthrough` are no longer skipped:

- `TLSRouteSimpleSameNamespace`
- `TLSRouteHostnameIntersection`
- `TLSRouteInvalidBackendRefNonexistent`
- `TLSRouteInvalidBackendRefUnknownKind`

`TLSRoute` is a CORE feature of the GATEWAY-TLS profile, so nothing is added to `SUPPORTED_EXTENDED_FEATURES`.

Their Gateway listener is fixed at port 443, and the suite dials the Gateway address there. One port cannot serve both HTTP-over-TLS and an untouched stream, so the conformance data plane points its 443 service port at APISIX's stream `tls_passthrough` listen instead of its HTTP ssl listen (`DeployDataplaneOptions.ServiceHTTPSTargetPort`). The HTTPRoute tests that would want HTTP-over-TLS on 443 — `HTTPRouteHTTPSListener`, `HTTPRouteRedirectPortAndScheme` — are already skipped for unrelated SAN reasons. The e2e data plane keeps 443 on the HTTP ssl listen and exposes the passthrough listen on 9120 of its own.

`TLSRouteModeMixed` is deliberately **not** claimed. Claiming it would require relaxing `portsWithConflictingTLSMode` and making the provisional `TLSRouteMixedTerminationSameNamespace` pass on a mixed listen; that is a follow-up.

## Tests

- `internal/adc/translator/tlsroute_test.go` — the mode of each port drives `tls_passthrough` (passthrough, terminate, omitted-mode, per-port, disagreeing listeners, portless, no listener); `sni` vs `snis`; the hostname fallbacks; per-port fan-out producing distinct ids.
- `test/e2e/gatewayapi/tlsroute.go` — a `Passthrough` listener on 9120 in front of the e2e nginx, with the client verifying the served chain against the **backend's** CA. The gateway holds no certificate for a Passthrough listener (it takes no `certificateRefs`), so a chain that validates there can only have come from the backend — which is what passthrough means.

`make lint` and `make test` are clean; `make generate manifests` produces no drift.

## Dependencies and verification status

- **Requires api7/adc#618.** ADC's `streamRouteSchema` is a `z.strictObject` and its Rust `StreamRoute` is `deny_unknown_fields`, so `tls_passthrough`/`snis` are rejected until that lands. `tls_passthrough` is only emitted for a `Passthrough` listener and `snis` only for more than one hostname, so a Terminate-only single-hostname setup — everything that works today — serializes byte-identically and is unaffected by the ADC version.
- `Makefile`'s `CONFORMANCE_DATAPLANE_VERSION` (3.17.0) predates apache/apisix#13912, so a **release** conformance report has to wait for the first APISIX release carrying it. A dev run picks it up from `apisix:dev` already; a comment now records that.
- The conformance suite has not been run end to end yet — it needs both the ADC change and an `apisix:dev` carrying #13912. The 443 remapping above is reasoned from the conformance manifests (no non-skipped test in the enabled profiles sends HTTPS traffic to 443), not yet observed.


AlinsRan:
Marked as draft: **CI cannot pass on this branch until api7/adc#618 is merged and a new `ghcr.io/api7/adc:dev` is published.**

The e2e and conformance jobs run against `adc:dev`, which is built from ADC `main`. ADC's `streamRouteSchema` is a `z.strictObject` (and its Rust `StreamRoute` is `deny_unknown_fields`), so `tls_passthrough` and `snis` are rejected outright rather than dropped — every TLSRoute Passthrough sync 400s. That chain is api7/api7ee-3-control-plane#3018 → api7/adc#618 → a published `adc:dev` → this PR.

To get evidence in the meantime, the conformance suite was run locally against an ADC image built from api7/adc#618 and `apache/apisix:master-debian-dev` (which carries apache/apisix#13912 — the released 3.17.0 does not). Results are in the PR description.

AlinsRan:
## Conformance run: green

Pushed two more commits and ran the suite end to end locally, against an ADC image built from api7/adc#618 and `apache/apisix:master-debian-dev` (which carries apache/apisix#13912; released 3.17.0 does not). `PROVIDER_TYPE=apisix`, experimental channel.

| Profile | Core | Extended |
|---|---|---|
| GATEWAY-HTTP | 32 passed, 0 failed, 5 skipped | 12 passed, 0 failed, 1 skipped |
| GATEWAY-GRPC | 15 passed, 0 failed, 0 skipped | 1 passed, 0 failed, 0 skipped |
| GATEWAY-TLS | **19 passed, 0 failed, 1 skipped** | 3 passed, 0 failed, 1 skipped |

Suite exits 0 and no profile reports `result: failure`, so the CI gate passes. GATEWAY-TLS core goes from 16 passed / 4 skipped on master to 19 passed / 1 skipped, and every skip on GATEWAY-HTTP was already there. The 443 remapping caused no regression, which was the part of the original description that had only been reasoned about rather than observed.

### What the run found

`TLSRouteSimpleSameNamespace`, `TLSRouteInvalidBackendRefNonexistent` and `TLSRouteInvalidBackendRefUnknownKind` passed on the first attempt — passthrough traffic works. `TLSRouteHostnameIntersection` did not, and produced two findings.

**1. A real defect, now fixed (`fix: intersect TLSRoute hostnames with the listener hostname`).**

A TLSRoute's hostnames were used verbatim as the SNI. Gateway API defines the effective hostnames as the intersection with the listener hostname. `tlsroute-more-specific-wc-hostname-x-1` carries `*.example.com` and attaches to a listener whose hostname is `abc.example.com`, so it was programmed as `sni: *.example.com` instead of `abc.example.com` — and since every Gateway shares one physical stream listen, it then answered `other.example.com`, which belonged to a sibling Gateway's route pointing at a different backend. That is what the test caught:

```
expected pod name to start with tls-backend-2, got tls-backend-75fb9c498c-4kf2x
```

HTTPRoute has narrowed its hostnames this way all along (`filterHostnames` + `getMinimumHostnameIntersection`); both now share `intersectRouteHostnames` and the TLSRoute reconciler applies it exactly as the HTTPRoute one does. Dumped from the data plane after the fix:

```
tlsroute-more-specific-wc-hostname-x-1   sni=abc.example.com   tls_passthrough=true
tlsroute-less-specific-wc-hostname-x-2   sni=*.example.com     tls_passthrough=true
tlsroute-less-specific-wc-hostname-x-4   sni=*.com             tls_passthrough=true
tlsroute-exact-hostname-x-{2,3,4}        sni=abc.example.com   tls_passthrough=true
```

Every intersection is now the smallest one, and no catch-all leaked.

**2. One assertion that cannot hold here, so the test is skipped as a known gap.**

`TLS request not matching hostnames intersection should not reach backend` expects a connection with SNI `non.matching.com` to the address of `gw-tlsroute-exact-hostname-x-1` to be rejected. `gw-tlsroute-empty-hostname-x-4`'s listener carries no hostname, so its route keeps `*.com` verbatim — correctly, and that Gateway's own subtest depends on it — and `*.com` matches `non.matching.com`. All four Gateways resolve to one data plane address and one stream listen, and which Gateway a connection was addressed to is not on the wire, so nothing is left to discriminate on.

It now sits next to `HTTPRouteMultipleGateways` in `skippedTestsForKnownGaps`, which is the same limitation, with that reasoning recorded. Every other assertion in the test passes, including the hostname intersections themselves.

### Still blocked

CI here stays red until api7/adc#618 is merged and a new `ghcr.io/api7/adc:dev` is published — see the previous comment.

AlinsRan:
api7/adc#618 is merged, so the schema side is done — but CI here is still blocked, on a **release** rather than the merge.

`kind-load-adc-image` pulls `ghcr.io/api7/adc:$(ADC_VERSION)` and retags it as `:dev`, so the published `adc:dev` is never what the e2e and conformance jobs run. `ADC_VERSION ?= 0.30.5`, and v0.30.5 was cut on 2026-09-16, before the merge:

```
$ docker run --rm -v ...:/w ghcr.io/api7/adc:0.30.5 validate -f /w/adc.yaml
✖ Unrecognized keys: "snis", "tls_passthrough"

$ docker run --rm -v ...:/w ghcr.io/api7/adc:dev validate -f /w/adc.yaml
(passes schema, proceeds to the backend call)
```

So this needs ADC to cut a release carrying api7/adc#618, and then a one-line `ADC_VERSION` bump here. Staying in draft until then; nothing else is outstanding — the conformance run is green locally against an ADC image built from that branch (GATEWAY-TLS core 19 passed / 0 failed / 1 skipped).

